PR #11, opened 2021-05-27
fix: disable interpretation of backslash escapes in response body

### LINKED ISSUE #10: parsing newlines
Hey, I'm running into some issues with newlines. For example when the response looks something like this:

```
{
    "a": "\n"
}
```

then I believe jq breaks somewhere along the way, since I get `parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 2, column 1`.

Otherwise a really neat project 😉

NTBBloodbath:
This PR disables the interpretation of backslash escapes in the response body rendering command. Should close #10

@rudotriton could you please test it to see if everything is alright?


rudotriton:
It does solve it, thank you 🙌

PR #13, opened 2021-05-27
fix: escape single quotes in response body

### LINKED ISSUE #12: problematic single quotes
I came across another thing. Something like this: a single single-quote

```json
{
  "contents": "'\n"
}
```

throws `zsh:1: unmatched "`

When there is an equal number of single quotes they just get removed in the result window. Doesn't seem to happen with double quotes or backticks.

NTBBloodbath:
This PR should fix the escaping of the single quotes in the response body. Should close #12

@rudotriton could you please test it to see if everything is alright? I didn't tested it, but should be working well.

### DIFF
--- a/lua/rest-nvim/init.lua
+++ b/lua/rest-nvim/init.lua
@@ -260,7 +260,7 @@ local function curl_cmd(opts)
 		if json_body then
 			-- Format JSON output and then add it into the buffer
 			-- line by line because Vim doesn't allow strings with newlines
-			local out = fn.system("echo -E '" .. string.gsub(line, "[']", "\\%1") .. "' | jq .")
+			local out = fn.system("jq", line)
 			for _, _line in ipairs(utils.split(out, '\n')) do
 				line_count = api.nvim_buf_line_count(res_bufnr) - 1
 				api.nvim_buf_set_lines(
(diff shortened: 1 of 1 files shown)


lua/rest-nvim/init.lua
```diff
@@ -260,7 +260,7 @@ local function curl_cmd(opts)
 		if json_body then
 			-- Format JSON output and then add it into the buffer
 			-- line by line because Vim doesn't allow strings with newlines
-			local out = fn.system("echo -E '" .. line .. "' | jq .")
+			local out = fn.system("echo -E '" .. string.gsub(line, "[']", "\\%1") .. "' | jq .")
```
disrupted, line 263: Instead of piping into stdin using echo I believe we could make use of the second argument to the system method
```suggestion
			local out = fn.system("jq", line)
```

disrupted (changes requested):

lua/rest-nvim/init.lua
disrupted: https://neovim.io/doc/user/eval.html#system()

rudotriton: well that is cool and it clears both of my recent issues

FWIW, `"\\%1"` wouldn't fix it anyway, but `"'\\%1'"` would

NTBBloodbath: Thank you for the suggestion, it makes sense! :)

PR #14, opened 2021-05-28
refactor: use lua table for response body

disrupted:
By splitting the response body to a table we should be able to use `nvim_buf_set_lines` in one call without having to iterate over all lines, which can take much longer. Also avoids some code duplication

Just gotta figure out how to run the tests, @NTBBloodbath to the rescue plz

### DIFF
--- a/lua/rest-nvim/init.lua
+++ b/lua/rest-nvim/init.lua
@@ -256,12 +256,12 @@ local function curl_cmd(opts)
 	end
 
 	--- Add the curl command results into the created buffer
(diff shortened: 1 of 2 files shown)


NTBBloodbath:
Hey thanks a lot, this will definitely improve performance and codebase! :)

To run the tests you only have to have `rest.nvim` installed and run it on the line where the request is (`METHOD URL`).

NTBBloodbath:
Sure, the truth is that a CI could be of great help. I'll add it soon :slightly_smiling_face:

NTBBloodbath (approved):

PR #24, opened 2021-06-30
refact: handle Authorization as a header

NTBBloodbath:
This PR fixes the `get_auth` function returning value so the `Authorization` header value will be successfully passed.

@mr-majkel can you please give it a try? I think that issue was by my side. Already passed an Authorization statement like `Authorization: Bearer random_characters` but I don't have how to properly test it right now, although I think it should work as intended.

Lemme know if the issue is fixed here and if that's the case then let's close the plenary issue regarding to Bearer tokens.

Regards

### DIFF
--- a/lua/rest-nvim/init.lua
+++ b/lua/rest-nvim/init.lua
@@ -200,7 +200,7 @@ end
 -- @param query_line Line to set cursor position
 local function get_auth(bufnr, query_line)
 	local auth = {}
-	local auth_not_empty = false
+	local is_auth_empty = true
 	-- Set stop at end of bufer
 	local stop_line = fn.line('$')
 
@@ -208,6 +208,10 @@ local function get_auth(bufnr, query_line)
 	for _ = 1, stop_line do
 		-- Case-insensitive search
 		local start_line = fn.search('\\cAuthorization:', '', stop_line)
+        if start_line ~= 0 then
+            is_auth_empty = false
+        end
+
 		local end_line = start_line
 		local auth_line = api.nvim_buf_get_lines(
 			bufnr,
@@ -227,7 +231,7 @@ local function get_auth(bufnr, query_line)
 	end
 
 	go_to_line(bufnr, query_line)
-	if not auth_not_empty then
+	if is_auth_empty then
 		return nil
 	end
 	return auth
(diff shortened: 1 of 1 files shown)


mr-majkel:
@NTBBloodbath indeed now the auth table is returned. However, with this request:

```http
POST https://reqres.in/api/users

Authorization: Bearer h3r3_is_sup3r10ng_t0ken!!!111
{
  "name": "morpheus",
  "job": "leader"
}
```

preview is showing we this:
```
[rest.nvim] Request preview:
curl -sSL -D /tmp/plenary_curl_2f4547a9.headers --compressed -X POST -H Content-Type:  application/json --data-raw {"name":"morpheus","job":"leader"} -u 1:Bearer https://reqres.in/api/users
```

so there is -u flag (username) that is assigned Bearer value. This seems to me plenary issue - where they are after user:password pair. What we are after with the token is the header passed to curl call:

```
-X Authorization: Bearer h3r3_is_sup3r10ng_t0ken!!!111
```

NTBBloodbath:
Alright, that's a plenary issue then, thanks to trying it out. I think I can do a workaround myself to fix it by passing the authentication header as a header instead of passing it as auth, do you want to try in that way?

mr-majkel:
Considering there is not so big movement on plenary side around my question - I think it's best approach.

I am also a bit confused on the auth field. Maybe the whole reason is that auth is supposed to be used only for -u option in curl, and rest of to-be-headers should be passed to headers table?

NTBBloodbath:
I think so, gimme a moment to modify the files and I'll let you know so you can try it

NTBBloodbath:
Alright, lemme know if it's working now!

mr-majkel:
@NTBBloodbath works like a charm:)

mr-majkel:
@NTBBloodbath I will ask you for yet another improvement related to this one:). I will make a separate issue for it, though, so we can discuss properly.

NTBBloodbath:
Thank you for testing it, I'm gonna merge it right now. Just remember to close the plenary issue because it was fixed by our side.

> I will ask you for yet another improvement related to this one:). I will make a separate issue for it, though, so we can discuss properly.

Sure, I'm gonna check it out right now, thank you!

Issue #31, opened 2021-07-20
[Feature Request] Import body from file


jens1205:
It would be nice if the body of the request could be imported from another file.

E.g.:
`POST http://localhost:8100/v1/orders
Accept: */*
Cache-Control: no-cache
Content-Type: application/json
X-Idempotency-Key: aaaaa

< ./order.json

###
`

jens1205:
See MR #32 for an implementation

NTBBloodbath:
Hey sorry for the late reply. Already merged it, thank you so much :slightly_smiling_face:

PR #32, opened 2021-07-20
allow import files for .http files

jens1205:
See #31 

import file directive with `< filename.xyz` will call curl with `-d @filename.xyz`

This works with absolute filenames and for relative filenames (where relative filenames will be interpreted as relative to the .http-file)


NTBBloodbath (approved):

NTBBloodbath:
Hey thank you so much for this new feature. I'm pretty sure that it's gonna be useful!

PR #35, opened 2021-07-21
Improvements

jens1205:
- fix broken import files (reverted unnecessary use of raw_body from yesterdays PR)
- fix broken header extraction
- better / more reliable support for multiple requests in one file
- jump to the request line of the chosen request
- curl error messages

### DIFF
--- a/lua/rest-nvim/init.lua
+++ b/lua/rest-nvim/init.lua
@@ -4,6 +4,7 @@ local curl = require('plenary.curl')
 local path = require('plenary.path')
 local utils = require('rest-nvim.utils')
 
+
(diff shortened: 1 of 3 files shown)


jens1205:
Hmm, I don't know selene, but for my understanding the code is correct.

Looks like your selene version does not support goto (https://github.com/Kampfkarren/selene/issues/224)

NTBBloodbath:
Hey, thank you so much, I'm really loving those changes!

Also regarding to selene linter it seems that we should change the stdlib to `lua52` for the `goto` statement, let me change it right now so we can merge this :)

PR #38, opened 2021-07-22
Dynamic variables

jens1205:
added support for dynamic variables 
    - $uuid: generates a universally unique identifier (UUID-v4)
    - $timestamp: generates the current UNIX timestamp (seconds since epoch)
    - $randomInt: generates a random integer between 0 and 1000
 
Documentation of dynamic variables (and also for import files)
Reorganized test-files, added content-type to examples.


jens1205:
@NTBBloodbath : Hmm, I thought your change in selene.toml to lua52 should fix this problem?

PR #44, opened 2021-08-07
Settings to open result on a horizontal split

samuelabreu:
As i split my buffers across vertical splits, it's better for me to have results on horizontal splits.


NTBBloodbath:
Hey, thank you so much for this. I'm pretty sure that others will find it helpful too :)

PR #45, opened 2021-08-07
Command RestNvimLast to repeat last request

samuelabreu:


lua/rest-nvim/init.lua
```diff
@@ -213,7 +216,6 @@ local function curl_cmd(opts)
     end
 
     local res_bufnr = get_or_create_buf()
-    local parsed_url = parse_url(vim.fn.getline('.'))
```
NTBBloodbath, line 216: Hey, why do you decided to remove this? That variable is very important because it encodes the URL to be able to pass arrays and such stuff and will also replace the environment variables if they exist in the URL.

This makes sense for me **only** when you're executing the last request because it will be already encoded :thinking:

NTBBloodbath (changes requested):

lua/rest-nvim/init.lua
samuelabreu: My mistake, sorry, didn't test it on different use cases.

I will fix it later. Thanks

NTBBloodbath: got it, no worries man!

Also you'll surely want to checkout the main branch again, I did some coding style conventions changes for having a more common style

samuelabreu: I updated to keep the current behavior when not re-executing a request.

Rewriting my code, i remembered i removed because `parse_url` was already being called on run() function, and didn't find any scenario where it needed to be revaluated, let me know if it's ok now.

Thank you

PR #50, opened 2021-08-16
feat: allow skipping SSL verification, closes #46

### LINKED ISSUE #46: Allow unknown certificate
Is there a way to force `curl -k` to skip ssl verification?

NTBBloodbath:
This PR aims to allow skipping SSL verification as the title says, this could be useful when dealing with unknown certificates.

To achieve this goal, the changes provides a new `setup()` function option called `skip_ssl_verification` which is `false` by default. If the option `skip_ssl_verification` is `true` then `rest.nvim` will pass the `-k` flag to cURL.

@rafi can you please test it out? :slightly_smiling_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,12 @@ use {
     'NTBBloodbath/rest.nvim',
     requires = { 'nvim-lua/plenary.nvim' },
     config = function()
-        require('rest-nvim').setup({
-          result_split_horizontal = false,
-        })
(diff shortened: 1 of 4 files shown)


rafi:
@NTBBloodbath works for me, thanks.

NTBBloodbath:
Thank you for using `rest.nvim`! :slightly_smiling_face:

Issue #56, opened 2021-08-24
External files: No variable substitution


jens1205:
Currently, there is no variable substitution for external files. 

In order to implement this, the original file cannot be passed to curl anymore. Perhaps the easiest way to implement is to read the file and do the rest as with local bodys.

Issue #55, opened 2021-08-24
External files import does not work any longer


jens1205:
Something  is broken with the external files.

E.g when trying the file `tests/external_file/post_create_user.http" the filename looks strange and is thus not recognized to be a file (which should be prefixed with @ in the curl call)

`curl -sSL -D /tmp/plenary_curl_47f448ac.headers --compressed -X POST -H Content-Type:  application/json --data-raw /./user.json https://reqres.in/api/users   `

jens1205:
see #56: When this ticket is implemented correctly, this issue is also solved

PR #57, opened 2021-08-24
replacement of variables also for imported files + refactoring (split into multiple modules)

jens1205:
This PR adresses #56 (and thus also #55)

It changes the following:
- Do not give curl filenames anymore. Instead read the files and apply variable substitution to all lines
- Fix filename resolution for relative filenames for importfiles
- refactor codebase, splitting the code into multiple modules
- simplify handling of last() function and LastOpts


jens1205:
Hi @NTBBloodbath ,

while implementing #56 I have also done some refactoring. I'd suggest to apply the four-eyes principle - do you mind to review the changes?

Regards

Jens

NTBBloodbath (approved):
Hey man, sorry for the delay, I was working haha

Thank you so much for this, being honest I haven't used the external bodies yet so I would never caught this bug by myself, nice catch!

I was also thinking since a long time to refact a ton of stuff here. I had a different idea about how things should be organized but this way also looks pretty good IMO.

Regarding to the four-eyes principle I didn't knew about it tbh so did a quick search on the web and the concept that I saw looks very good, I think that we should definitely use it, in that way we can improve a ton of stuff, including the transparency and that's a really good point. Let me know how can we start implementing this :)

Regards

jens1205:
If this would be a Greenfield project my structure would also look a bit different. But this first refactoring was a trade off between better readability / maintainability ob one side and effort/risk on the other.

We could improve based on this. But for a comparable small project like this there is no strong need to have the best possible solution. 

Thanks for the review!

NTBBloodbath:
Yeah, agreed. The code structure was pretty much of junk before, hopefully this refactoring will open the doors for future contributors and obviously for us too.

And no problem man, thank you for doing this!

PR #60, opened 2021-08-30
Highlight request on run() and last()

jens1205:
New feature "highlight request on run() and last()". In order to implement this, we had to switch from synchronous call to plenary-curl to asynchronous with callback. Implementation is similar to the build-in `vim.highlight.on_yank`. This feature is enabled by default.

Additionally, the old feature of moving the cursor to the current request line is disabled by default as the highlighting shows more accurate what is send and in my opinion it is nicer to have the cursor stay in place.

@NTBBloodbath : Any objections?

### DIFF
--- a/lua/rest-nvim/curl/init.lua
+++ b/lua/rest-nvim/curl/init.lua
@@ -41,69 +41,68 @@ M.get_or_create_buf = function()
 end
 
 local function create_callback(method, url)
-   return function(res)
-      if res.exit ~= 0 then
-        log.error("[rest.nvim] " .. utils.curl_error(res.exit))
-        return
-      end
-      local res_bufnr = M.get_or_create_buf()
-      local json_body = false
+  return function(res)
+    if res.exit ~= 0 then
+      log.error("[rest.nvim] " .. utils.curl_error(res.exit))
+      return
+    end
+    local res_bufnr = M.get_or_create_buf()
+    local json_body = false
 
-      -- Check if the content-type is "application/json" so we can format the JSON
-      -- output later
-      for _, header in ipairs(res.headers) do
-        if string.find(header, "application/json") then
-          json_body = true
-          break
-        end
+    -- Check if the content-type is "application/json" so we can format the JSON
+    -- output later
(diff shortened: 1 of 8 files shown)


lua/rest-nvim/init.lua
```diff
@@ -25,8 +25,15 @@ rest.run = function(verbose)
     raw = config.skip_ssl_verification and { "-k" } or nil,
     body = result.body,
     dry_run = verbose or false,
+    bufnr = result.bufnr,
+    start_line = result.start_line,
+    end_line = result.end_line,
   }
 
+  if config.get("highlight").enabled == true then
```
NTBBloodbath, line 33: We can skip the `== true` here, it's not really needed :p

lua/rest-nvim/curl/init.lua
```diff
@@ -39,75 +40,83 @@ M.get_or_create_buf = function()
...
-    res_bufnr,
-    line_count + 1,
-    line_count + 1 + #res.headers,
-    false,
-    res.headers
-  )
+    -- Only open a new split if the buffer is not loaded into the current window
+    if vim.fn.bufwinnr(res_bufnr) == -1 then
+      local cmd_split = [[vert sb]]
+      if config.result_split_horizontal == true then
```
NTBBloodbath, line 95: we can skip the true verification here too :p

NTBBloodbath (changes requested):
Looking really good man, thank you for doing this!

I don't have any objection haha, but found some checks to `== true` that makes no sense imo since the conditionals implicitly do this but didn't wanted to be an ass and request a ton of changes for all of them so you know, just change it and feel free to merge. Awesome job man, thanks for contributing!

lua/rest-nvim/init.lua
jens1205: ok

lua/rest-nvim/curl/init.lua
jens1205: ok

NTBBloodbath (approved):

PR #63, opened 2021-09-06
Load .env files in dir of currently opened file

jim-fx:
Hi @NTBBloodbath,

This pr adds the loading of .env files in the directory of the currently opened file.

## Usecase
I have a big monorepo structured like

```
├─ backend
│  └─ requests
│        └─  test.http
└─ frontend
```
 
Currently, I have to place my .env file for common properties in all my .http files in the root folder of my project.
I would like to place them in the same folder where all my requests are.

## Considerations

**Which variables overwrite which?**

Probably .env files in deeper nested directories should overwrite those in upper directories.

**Should .env files in intermediate directories also be loaded?**

Should it also load .env files in `/backend/.env`? Probably it should if that doesn't add a lot of overhead.

Thank you a lot for this awesome plugin @NTBBloodbath,

Cheers, Jim

Edit: 
Sorry for all the fix commits, i don't have a lot of experiences with pr's/lua

### DIFF
--- a/lua/rest-nvim/utils/init.lua
+++ b/lua/rest-nvim/utils/init.lua
@@ -49,12 +49,22 @@ end
 -- returns a table with the variables
 M.read_env_file = function()
   local variables = {}
-  local env_file_path = vim.fn.getcwd() .. "/.env"
-  -- If there's an env file in the current working dir
-  if M.file_exists(env_file_path) then
-    for line in io.lines(env_file_path) do
-      local vars = M.split(line, "=", 1)
-      variables[vars[1]] = vars[2]
+  
+	-- Directories to search for env files
+  local env_file_paths = {
+		-- current working directory 
+		vim.fn.getcwd() .. "/.env", 
+		-- directory of the currently opened file
+		vim.fn.expand("%:p:h") .. "/.env"
+	}
+
+	-- If there's an env file in the current working dir
+  for env_file_path in env_file_paths do
+    if M.file_exists(env_file_path) then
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):

NTBBloodbath:
Hi, thank you so much for doing this, it's something that I really wanted to do since a long time ago.

> **Which variables overwrite which?**
> 
> Probably .env files in deeper nested directories should overwrite those in upper directories.

Yeah I think so, they're gonna be overwritten by precedence.

> **Should .env files in intermediate directories also be loaded?**
> 
> Should it also load .env files in `/backend/.env`? Probably it should if that doesn't add a lot of overhead.

I don't see why not, someone will surely have an use case for it haha, I don't think that it would affect too much the performance, afaik the environment files aren't too big.

> Thank you a lot for this awesome plugin @NTBBloodbath

Thank you for using it and also for helping to improve it!

> Sorry for all the fix commits, i don't have a lot of experiences with pr's/lua

No problem haha we can squash them to convert all your commits into a single commit when merging :slightly_smiling_face: 

---

Just a small thing before merging, our linter CI failed in the PR, you can see why by clicking the `Details` link.

Regards

jim-fx:
Hi, 

Thanks for the detailed reply :)

I installed luacheck and fixed the unused variable error, i think the linter should be happy now.

Cheers, Jim

NTBBloodbath (approved):

NTBBloodbath:
Hey, thanks for fixing it, awesome job man!

PR #81, opened 2021-11-13
Allow hiding specific parts of the result like url, headers, and http info.

khalidchawtany:
Sometimes I just need the result and don't care bout other parts. I don't want to scroll to get to the response body.


NTBBloodbath (approved):
Thank you so much per doing this :)

PR #86, opened 2021-11-29
Document variables

### LINKED ISSUE #68: [Feature request] Support mode for env variables.
Nice work! This is the restclient I'm looking for a long time.

Here is a common case when I test restful api in day to day work. 

Define a env named {{base_url}} in both `production` and `local` modes. So I could switch to `local` to test the code in local server, and switch to `prod` after deployment.

The {{base_url}} may looks like this.

In `production` mode, base_url is `http://xyz.abc/api/v1.0/something`, and in `local` mode base_url is `http://localhost:8080`.

If rest.nvim could support this, I could replace my rest client for good.

klen:
Found this plugin and it's awesome thanks! I was missing fast switching of variables in the document, so I made this feature.
I suppose this will fix #68.

Just an example:
```
# Some API

# Production settings
host: "service.com"
token:  "PRODUCTION_TOKEN"
username: "manager"

# Stage settings
# host: "stage.service.com"
# token:  "STAGE_TOKEN"
# username: "admin"

query: 'where={"query":"test"}'

###

# Get users
GET https://{{host}}/api/users?{{query}}
Authorization: Bearer {{token}}

# Create a user
POST https://{{host}}/api/users
Authorization: Bearer {{token}}

{
    "name": "{{username}}"
}

```
You may fast change/comment/uncomment the variables you need and keep it with your `http` document.

### DIFF
--- a/lua/rest-nvim/utils/init.lua
+++ b/lua/rest-nvim/utils/init.lua
@@ -92,7 +92,7 @@ end
 
 M.get_node_value = function(node, bufnr)
   local start_row, start_col, _, end_col = node:range()
-  local line = api.nvim_buf_get_lines(bufnr, start_row, start_row + 1, false)[1]
+  local line = vim.api.nvim_buf_get_lines(bufnr, start_row, start_row + 1, false)[1]
   return line and string.sub(line, start_col + 1, end_col):gsub('^["\'](.*)["\']$', '%1') or nil
 end
 
--- a/lua/rest-nvim/utils/init.lua
+++ b/lua/rest-nvim/utils/init.lua
@@ -90,15 +90,44 @@ M.read_dynamic_variables = function()
   return dynamic_variables
 end
 
+M.get_node_value = function(node, bufnr)
+  local start_row, start_col, _, end_col = node:range()
+  local line = api.nvim_buf_get_lines(bufnr, start_row, start_row + 1, false)[1]
+  return line and string.sub(line, start_col + 1, end_col):gsub('"', '') or nil
+end
+
+M.read_document_variables = function()
+  local variables = {}
+  local bufnr = vim.api.nvim_get_current_buf()
+  local parser = vim.treesitter.get_parser(bufnr)
+  if not parser then return variables end
+
+  local first_tree = parser:trees()[1]
+  if not first_tree then return variables end
+
+  local root = first_tree:root()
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):

NTBBloodbath:
Hi, looks great to me (and is even more simple that the solution that I thought :laughing:). Thank you so much!

lua/rest-nvim/utils/init.lua
```diff
@@ -90,15 +90,44 @@ M.read_dynamic_variables = function()
...
+  local first_tree = parser:trees()[1]
+  if not first_tree then return variables end
+
+  local root = first_tree:root()
+  if not root then return variables end
+
+  for node in root:iter_children() do
+    local type = node:type()
+    if type == 'header' then
+      local name = node:named_child(0)
```
dakennguyen, line 114: Hi, thank you for the implementation. 
I see that this PR use node type `header` for local variable section, but treesitter `header`'s name have this pattern `[A-Za-z-]+` ([here](https://github.com/NTBBloodbath/tree-sitter-http/blob/main/src/grammar.json#L722-L725)). It means that we can't use underscore for the variable name, and consider snake_case or UPPERCASE naming convention, I think underscore is necessary
image

Would it be possible to add a new treesitter group for local variables, or if we reuse `header` pattern, could we add underscore?

NTBBloodbath: Hi, can you please create an issue in the tree-sitter parser repo? In that way we could have a better discussion about what to do. The parser still needs a bit more of work to work in a 1:1 way with rest.nvim features so opening issues in the repo would be great IMO.

Also thank you per ponting this out, I completely missed this :)

klen: There could be some solutions:

1. Right now you can use CamelCase
```http
contributionID: 456

GET https://test.com/?{{contributionID}}
```
2. We can fix `rest-nvim.utils` to accept "dashed" var names, like this (I may provide PR, just a line of code):
```http
contribution-id: 456

GET https://test.com/?{{contribution-id}}
```
3. We can replace "dashed" var names to underscore, like this (also the small change) btw it's not explicit:
```http
contribution-id: 456

GET https://test.com/?{{contribution_id}}
```

PR #87, opened 2021-12-09
Allow spaces in variable names, and  cross compatibility with $dotenv

Matsuuu:
The VSCode REST client uses a {{$dotenv URL}} pattern to match .env file variables. I work in a team where not everyone uses neovim, so having a cross compatiblity layer for this would be good.

The PR changes how the url is parsed, allowing for spaces in the path and then removing the dotenv references, so that the same markdown works with this client

Example case:

```
GET {{$dotenv URL}}/test/auth
```

### DIFF
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -161,10 +161,14 @@ end
 -- @param stmt the request statement, e.g., POST http://localhost:3000/foo
 local function parse_url(stmt)
   local parsed = utils.split(stmt, " ")
+  local http_method = parsed[1]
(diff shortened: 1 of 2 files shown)


Matsuuu:
Friendly ping @NTBBloodbath .

This is not going to affect the performance of the tool, but would allow for some interop between tooling

NTBBloodbath:
Hi, thank you so much per doing this, loving it so far:)

Also thank you so much per doing that ping haha I completely forgot this PR with my other projects and work so my apologies!

Reviewing your awesome work right now before merging :)

NTBBloodbath (approved):

PR #88, opened 2021-12-14
fix: use config.get to read skip_ssl_verification setting

joshuali925:
Thanks for the plugin. I believe the code was refactored at some point and `config` currently only has `.get()` and `.set()` functions. `skip_ssl_verification` worked again for me after this fix.


NTBBloodbath:
Woops, nice catch! I wonder how did we forgot this, thank you so much :)

PR #90, opened 2021-12-16
feat: yank executable curl cmd

### LINKED ISSUE #71: Command/binding to copy request as curl
Would be nice to share with other people as a curl command, and not just printing it for inspection.
I noticed that the curl command is including some headers from plenary. Would it be posible to keep this to the simplest for this functionality? 
Thanks! :)

purefun:
The pr may close #71


NTBBloodbath (approved):
Great, thank you so much! Very appreciated :heart:

PR #93, opened 2021-12-31
make RestNvimLast re-run the last run query and not a preview

### LINKED ISSUE #92: RestNvimLast re-runs Preview if it was the last issued command to rest nvim
Maybe this is the intended behaviour?

I have this to test:

```http
GET https://httpbin.org/get?1=1
GET https://httpbin.org/get?2=2
```
If i do RestNvim on the first line, i get the answer for the first query.

I move to line 2 and do RestNvimPreview, it shows the curl request at the bottom of my screen.

I then do RestNvimLast and it shows the preview again.

I was expecting it to re-run the first query, but that's definitely something i can live with ;)

gtnbssn:
fixes #92

### DIFF
--- a/lua/rest-nvim/init.lua
+++ b/lua/rest-nvim/init.lua
@@ -2,6 +2,7 @@ local rest = {}
 local request = require("rest-nvim.request")
 local config = require("rest-nvim.config")
 local curl = require("rest-nvim.curl")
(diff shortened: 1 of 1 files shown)


NTBBloodbath:
Thank you so much!

PR #120, opened 2022-06-23
fix: pass the body as string regardless of it is json

teto:
fix: pass the body as string regardless of it is json   else we hit errors in plenary.

In Plenary we can set the "body" as a string (either filename or the actual content of the payload)
or as a table in which case it is passed as `curl --data` . According to the doc, the latter does:
```
        -d, --data <data>
                  (HTTP MQTT) Sends the specified data in a POST request to the HTTP server, in the same way that a browser does when a user has filled in an HTML form and presses the submit button. This will cause curl to pass the data to the server using the content-type application/x-www-form-urlencoded. Compare
                  to -F, --form.
```
Currently rest.nvim sends a table when the file is a json one which makes no sense as per curl doc so always send it as a string (aka as "content")

    The relevant plenary code;
    https://github.com/nvim-lua/plenary.nvim/blob/1da13add868968802157a0234136d5b1fbc34dfe/lua/plenary/curl.lua#L183-L197

### DIFF
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -71,11 +71,6 @@ local function get_body(bufnr, start_line, stop_line)
     end
   end
 
-  local is_json, json_body = pcall(vim.fn.json_decode, body)
-  if is_json then
-    return json_body
-  end
-
   return body
 end
 -- is_request_line checks if the given line is a http request line according to RFC 2616


teto:
btw looking at https://www.reddit.com/r/neovim/comments/ug9ypl/restnvim_is_looking_for_maintainerscontributors/ , I dont really want to get involved with the development but I wouldn't mind merging rights to be able to merge the occasionnal fixes.

teto:
It's possible to see the issue using the following json file 
```
{
    "_categoricalParams": [
        {
            "allLevels": [
                "one",
                "three",
                "two"
            ],
            "currentLevel": "one",
            "description": "",
            "name": "cat_param1",
            "reference": ""
        },
        {
            "allLevels": [
                "one",
                "three",
                "two"
            ],
            "currentLevel": "one",
            "description": "",
            "name": "cat_param2",
            "reference": ""
        },
        {
            "allLevels": [
                "one",
                "three",
                "two"
            ],
            "currentLevel": "one",
            "description": "",
            "name": "cat_param3",
            "reference": ""
        }
    ]
	}
	```

I get a 
```
Traceback: /home/teto/neovim/plenary.nvim/lua/plenary/curl.lua:57: attempt to concatenate a table value  
```

teto:
apparently fixed by https://github.com/rest-nvim/rest.nvim/pull/125 (haven't tested)

PR #125, opened 2022-08-07
Quality of life changes to rest.nvim

### LINKED ISSUE #67: Error parsing bug
Hello. I faced up with parsing error bug.
Look at the response of the `rest.nvim`
image
and now look at the response of the same request in insomnia:
image

### LINKED ISSUE #114: Sending JSON-payload
Hello! I just installed this plugin, and it works pretty well so far! I'm having some issues though, not sure if I'm understanding this plugin correctly. I'm trying to send some data as json, like this:

```
POST http://httpbin.org/post

{ "name": "Test"}
```

I get a response like this in the response window:
```
POST http://httpbin.org/post

{
  "args": {},
  "data": "name=Test",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "*/*",
    "Accept-Encoding": "deflate, gzip, br",
    "Content-Length": "9",
    "Content-Type": "application/json",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.68.0",
  },
  "json": null,
  "origin": "XXX.XXX.XXX.XXX",
  "url": "http://httpbin.org/post"
}

```

I was expecting to get the same data back in the `json` -field of the response, but it's `null` as you can see. I checked the output of `RestNvimPreview` and it gave me this: ` curl -sSL --compressed -X 'POST' -H 'Content-Type: application/json' -d 'name=Test' 'http://httpbin.org/post'` . Trying to send that from the command line gives me the same response as before. 

I kinda figured with the `Content-Type` header, the data would be sent as json, but it seems like that's not the case. Is this intended behavior, or have I misunderstood something? It's a great plugin nonetheless, thanks for creating it!

### LINKED ISSUE #118: plenary/curl : attempt to concatenate a table value
I have an error when there is a list in somem field. The test in the repository doesn't pass : https://github.com/NTBBloodbath/rest.nvim/blob/main/tests/post_create_user.http

the error message is : 
```
Traceback: ...site/pack/packer/start/plenary.nvim/lua/plenary/curl.lua:54: attempt to concatenate a table value
```
NeoVim version : v0.7.0
packer :   
```
use { 'NTBBloodbath/rest.nvim',
    requires = { "nvim-lua/plenary.nvim" },
    config = function() require('configure.restvim') end
  }
```
configure.restvim : 

```
require("rest-nvim").setup({
     -- Open request results in a horizontal split
    result_split_horizontal = false,
    -- Keep the http file buffer above|left when split horizontal|vertical
    result_split_in_place = false,
    -- Skip SSL verification, useful for unknown certificates
    skip_ssl_verification = false,
    -- Highlight request on run
    highlight = {
      enabled = true,
      timeout = 150,
    },
    result = {
      -- toggle showing URL, HTTP info, headers at top the of result window
      show_url = true,
      show_http_info = true,
      show_headers = true,
    },
    -- Jump to request line on run
    jump_to_request = false,
    env_file = '.httpenv',
    custom_dynamic_variables = {},
    yank_dry_run = true,
})

vimp = require('vimp')
vimp.nnoremap('<leader>œ', '<Plug>RestNvim')
```

### LINKED ISSUE #123: Question: How can I pass arguments to curl ?
I would like to add `--location` argument to curl. Is this possible using rest.nvim ?

udayvir-singh:
#### Add ability to add custom formatters.
setup function now accepts `formatters` table. this makes `jq` optional.

```lua
require [[rest-nvim]].setup {
  result = {
    -- executables for formatting response body [optional]
    formatters = {
      -- set them to nil if you want to disable them
      json = "jq",
      html = {"tidy", "-i", "-q", "--show-errors", "0", "-"}
    },
  }
}
```

#### Add better syntax highlights in httpResult.
- properly highlighted http status codes.
- add separate highlights for date/expires headers.
- better regex for numbers.
- support highlighting response based on content-type header.

highlights

#### Add support for passing flags to curl, fixes #123.

Example:
```js
POST http://localhost:3000/

-c /tmp/cookie

{ "id": 1 }
```

#### Fix #67 error handling for formatters.


README.md
```diff
@@ -47,7 +47,6 @@ have to leave Neovim!
 
 - System-wide
   - curl
-  - jq +1.6 (to format JSON output so it can be human-readable)
```
NTBBloodbath, line 50: Maybe keep it as an optional dependency?

Also I think tidy should be added as an optional dependency too as AFAIK it is an external tool right?

lua/rest-nvim/request/init.lua
```diff
@@ -42,7 +42,8 @@ end
 -- @param bufnr Buffer number, a.k.a id
 -- @param start_line Line where body starts
 -- @param stop_line Line where body stops
-local function get_body(bufnr, start_line, stop_line)
+-- @param has_hson True if content-type is set to json
```
NTBBloodbath, line 45: `has_json` parameter name has a small typo here :p

NTBBloodbath (commented):
Hey, thank you so much for this!

All these changes looks great, just a small things to be fixed (only documentation) imho.

I'll hopefully merge this PR once these changes are made :)

Cheers

README.md
udayvir-singh: @NTBBloodbath

`jq` and `tidy` can be replaced by other executable or even native lua functions, that's why I didn't add them as dependencies.

udayvir-singh: I don't think they should be added as optional dependencies as people would just install them thinking they are required.

NTBBloodbath: Ah, makes sense now. Yeah let's leave it as this then, thanks for explaining

NTBBloodbath (approved):

NTBBloodbath (approved):

lua/rest-nvim/curl/init.lua
```diff
@@ -114,6 +147,19 @@ local function create_callback(method, url)
 
     -- Send cursor in response buffer to start
     utils.move_cursor(res_bufnr, 1)
+
+    -- add syntax highlights for response
+    local syntax_file = vim.fn.expand(string.format("$VIMRUNTIME/syntax/%s.vim", content_type))
+
+    if vim.fn.filereadable(syntax_file) == 1 then
+      vim.cmd(string.gsub([[
+        unlet b:current_syntax
```
teto, line 156: this line triggers an error "unknown b:current_syntax" if I run a request twice

NTBBloodbath: I did push a small fix for this that checks if `b:current_syntax` exists a few seconds ago, please let me know if the issue persists.

teto: seems like that fixed it thx

teto:
I've been trying to override/remove the default formatters since otherwise I get 
```
Error calling formatter on response body:                                                                                                                                                                                                                       
Vim:E475: Invalid value for argument cmd: 'tidy' is not executable
```

so I've done:
```
require("rest-nvim").setup({
          -- Open request results in a horizontal split
          result_split_horizontal = false,
          -- Skip SSL verification, useful for unknown certificates
          skip_ssl_verification = false,
          -- Highlight request on run
          highlight = {
		   enabled = true,
		   timeout = 150,
          },
          result = {
		   -- toggle showing URL, HTTP info, headers at top the of result window
		   show_url = true,
		   show_http_info = true,
		   show_headers = true,
		   -- disable formatters else they generate errors/add dependencies
		   -- for instance when it detects html, it tried to run 'tidy'
		   formatters = {
			html = nil,
			jq = nil
		   },
          },
          -- Jump to request line on run
          jump_to_request = false,
		})
```
to no avail. Can you change/override the formatters ? if yes how ? if no, do you want to send a patch ?

udayvir-singh:
@teto

I just noticed you have to set formatter to `false` instead of `nil` to disable them, also I have added an extra check to disable html formatter if tidy is not available in PR #165, and fixed the README in it.

teto:
do you know why we special case json when we send the payload ? there should be no need, shouldn't it ? 

I am looking at https://github.com/rest-nvim/rest.nvim/issues/157 and so if we didn't have to special case json, then we could skip the content-type check.

PR #127, opened 2022-08-09
feat: adding rockspec

teto:
so that one can install the plugin from luarocks.org .
It helps deal automatically with plugin dependencies.

See https://teto.github.io/posts/2022-06-22-neovim-plugin-luarocks-2.html for more context.

### DIFF
--- a/rest.nvim-scm-1.rockspec
+++ b/rest.nvim-scm-1.rockspec
@@ -1,0 +1,36 @@
+local MAJOR, REV = "scm", "-1"
+rockspec_format = "3.0"
+package = "rest.nvim"
+version = MAJOR .. REV
+
+description = {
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):

NTBBloodbath:
Thank you so much!

I think this could also help us later to move from plenary's cURL wrapper and use a library like lua-cURL or lluv-curl (that one is an async lib using libuv and libcurl) from luarocks for a more solid and feature-complete cURL wrapper/bindings.

teto:
arf I forgot to add plenary as a dependency :'(

NTBBloodbath:
Woops, didn't notice at first (just woke up haha). Gotta make a new PR ig :p

PR #128, opened 2022-08-09
adds a Makefile

teto:
makes the available actions more discoverable.
Ideally we shouldn't have to push and wait for CI to find issues with a
PR.


NTBBloodbath (approved):
Thank you so much! :)

PR #129, opened 2022-08-09
rockspec: add missing dependency plenary

teto:
this wont work because I had to remove the plenary package from luarocks but it's only temporary: https://github.com/nvim-lua/plenary.nvim/pull/391
should be fine in a couple days.


NTBBloodbath (approved):

teto:
I've uploaded it here https://luarocks.org/modules/teto/rest.nvim

PR #131, opened 2022-08-10
refactor file_exists function

Yagua:
Since this implementation of the `file_exists` function has the ability to give an answer as to whether or not a file exists, it is not completely accurate since giving the `open` function a directory path as an argument will still return `true` indicating that the file exists even though it is not actually a file.

```lua
M.file_exists = function(file)
  file = io.open(file, "rb")
  if file then
    file:close()
  end
  return file ~= nil
end
````

so I refactored the function, making use of the built-in function`'vim.fn.filereadable`, to ensure that first we are verifying that we are actually reading a file and not a directory, and in turn showing if this file exists or not.

```lua
M.file_exists = function(file)
  if vim.fn.filereadable(file) ~= 1 then return false end
  return true
end
```

In this way when running the `read_env_file` function and there is a virtual environment (directory) with the name `.env` in the current directory, only the `.env` files will be taken into account and not the directory, preventing `io.lines` from trying to scan a directory.

### DIFF
--- a/lua/rest-nvim/utils/init.lua
+++ b/lua/rest-nvim/utils/init.lua
@@ -27,11 +27,8 @@ end
 -- file_exists checks if the provided file exists and returns a boolean
 -- @param file File to check
 M.file_exists = function(file)
-  file = io.open(file, "rb")
-  if file then
-    file:close()
-  end
-  return file ~= nil
+  if vim.fn.filereadable(file) ~= 1 then return false end
+  return true
 end
 
 -- read_file Reads all lines from a file and returns the content as a table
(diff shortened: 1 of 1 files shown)


lua/rest-nvim/utils/init.lua
```diff
@@ -27,11 +27,8 @@ end
 -- file_exists checks if the provided file exists and returns a boolean
 -- @param file File to check
 M.file_exists = function(file)
-  file = io.open(file, "rb")
-  if file then
-    file:close()
-  end
-  return file ~= nil
+  if vim.fn.filereadable(file) ~= 1 then return false end
+  return true
```
NTBBloodbath, line 31: Maybe we could simplify that with something like this

```suggestion
  return vim.fn.filereadable(file) == 1
```

NTBBloodbath (changes requested):
Hey,

This is indeed a great catch, thanks for the refactoring!

I would like the change I suggested to be applied before merging tho, I feel like it's worth it to simplify codebase and improve its readability.

Cheers

Yagua:
Of course, it looks even better!

NTBBloodbath (approved):

PR #138, opened 2022-08-28
fix parsing of nested tables and curl arguments

udayvir-singh:
Fixes errors when you pass array or object in request body without setting request body to `application/json`, for example:

```http
POST http://localhost:8000/login

{
  "username": "foo",
  "args": {
    "key": "val"
  }
}
```

would result in following error:
```
Traceback: ...site/pack/packer/start/plenary.nvim/lua/plenary/curl.lua:54: attempt to concatenate a table value
```

but this PR fixes it and gives the following query:
```bash
curl -sSL --compressed -X 'POST' -d 'args={"key": "val"}' -d 'username=foo' 'http://localhost:8000/login'
```

### DIFF
--- a/lua/rest-nvim/config/init.lua
+++ b/lua/rest-nvim/config/init.lua
@@ -15,22 +15,15 @@ local config = {
     formatters = {
       json = "jq",
       html = function(body)
-        return vim.fn
-          .system({
-            "tidy",
-            "-i",
-            "-q",
-            "--tidy-mark",
-            "no",
-            "--show-body-only",
-            "auto",
-            "--show-errors",
-            "0",
-            "--show-warnings",
-            "0",
-            "-",
-          }, body)
-          :gsub("\n$", "")
+        -- stylua: ignore
+        return vim.fn.system({
+          "tidy", "-i", "-q",
+          "--tidy-mark",      "no",
+          "--show-body-only", "auto",
+          "--show-errors",    "0",
+          "--show-warnings",  "0",
+          "-",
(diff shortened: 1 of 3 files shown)


NTBBloodbath:
Hey @teto, hope you're doing well (and sorry for the mention).

Do you think it's worth it merging this one? I was checking the PRs that are open and found #120 (that is yours) and it does fix body send by sending the whole body as a string.

Would your PR fix this problem and many others too? If so, we could update your PR branch and merge that one instead and close this one in favor of #120.

udayvir-singh:
@NTBBloodbath the issue that #120 was trying to solve has already been fixed by my previous PR #125.

This PR fixes errors when you try to include json values in `application/x-www-form-urlencoded` content type which is default on curl.

teto:
I was wondering if my PR was still relevant, thanks for addressing this. Could you add a test that fails without this please ?

udayvir-singh:
@teto

I have added a test `post_json_form`, it should return a error if you try to run it from the master branch.

NTBBloodbath (approved):

NTBBloodbath:
Hey, sorry for the late reply!

Thought actual codebase rewrite would take less time (it's still in its time to wear diapers, but I'm making very good decisions behind the scenes!) so I didn't want to merge any actual PR in order to keep codebase simpler to track for me, however, that's not the case right now and we gotta merge these fixes _and some new features_ as people really needs them.

Also thank you so much for this PR changes, everything looks good to me. Thank you again for fixing cURL arguments parsing too!

PR #140, opened 2022-09-17
Fix indentation for new `encode_url` option.

rhart92:
Swap tabs to spaces so everything is aligned. 😅 

CleanShot 2022-09-16 at 17 06 40@2x


NTBBloodbath:
Thank you and sorry for the delay

Issue #141, opened 2022-10-11
Touches names of Header values


appelgriebsch:
My Rest API requires me to specify a header value of `TOKEN: <token>` with the name of the header value in all-upper-case. When executing the request the API always complains of not having a token in the header.... checking the cUrl command that is issued it seems the plugin changes the the header name from all upper-case to pascal case...

```
GET https://test/me
**TOKEN**: {{TOKEN}}
Accepts: application/json
```

result in this cUrl:

```
curl -sSL --compressed -X 'GET' -H '**Token**: ...' -H 'Accepts: application/json' 'https://test/me'
```

appelgriebsch:
If it’s related to this line: https://github.com/rest-nvim/rest.nvim/blob/main/lua/rest-nvim/request/init.lua#L124 I might be able to provide a PR for it tomorrow…

appelgriebsch:
Ok, made a small local improvement on the above mentioned code line. But stumbled upon another one:

If you specify `API_TOKEN: <token>` as header the cUrl command will send a `API-TOKEN: <token>` instead. This was a bit harder to figure out, but looks like its an issue in the plenary: https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/curl.lua#L95

NTBBloodbath:
Reopened this issue until rest.nvim codebase rewrite is done as there's another bug you've mentioned that is related to plenary _and codebase rewrite should be able to fix this behavior_.

rhinoxi:
Hi @NTBBloodbath , it seems PR https://github.com/rest-nvim/rest.nvim/pull/142 broke `Content-Type: application/json` header, only `content-type: application/json` works fine now.

teto:
I dont know why we special case json, it justs decode then reencode the content :s 
If i remove the special casing it just works !
```
diff --git a/lua/rest-nvim/request/init.lua b/lua/rest-nvim/request/init.lua
index 9d7727b..905efd7 100644
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -41,8 +41,7 @@ end
 -- @param bufnr Buffer number, a.k.a id
 -- @param start_line Line where body starts
 -- @param stop_line Line where body stops
--- @param has_json True if content-type is set to json
-local function get_body(bufnr, start_line, stop_line, has_json)
+local function get_body(bufnr, start_line, stop_line)
   -- first check if the body should be imported from an external file
[...]

PR #142, opened 2022-10-13
#141: fixes header name

appelgriebsch:
Instead of converting the captured header name to lowercase using it the way it is spelled in the document.


NTBBloodbath (approved):
Thank you so much for this PR!

teto:
as mentioned in the related PRs, this breaks rest-nvim so what was the motivation in the first place ?

rochacbruno:
I spent some good minutes trying to figure out what was wrong with my API code and it was just this.

Maybe add to the docs.

**Headers must be lowercase**
```diff
- Content-Type: application/json
+ content-type: application/json
```

Which makes no sense, HTTP std accepts capital headers.

PR #148, opened 2022-11-10
Added in variable substituion for environment variables, and made them be able to be local to the file

theNEOkun:
Added the possibilty to start a variable with @ in the local file, and then also to use substitutions in the variables, I.E if

@source=api
@port=3000
@path=http://localhost
@url={{path}:{{port}}/{{source}}

then url = http://localhost:3000/api


teto:
> Added the possibilty to start a variable with @

any reason why this is interesting compared to `{{ }}` ? 

> nd then also to use substitutions in the variables, I.E if

I am worried about the complexity of this.

Anyhow can you clean up the history ?

Issue #132, opened 2022-08-12
File variable support


mawkler:
VSC**es's REST client plugin, which is similar to rest.nvim, supports [file variables](https://github.com/Huachao/vscode-restclient#file-variables). Here's an example to illustrate:

```
@baseUrl = https://test.url.com
@access_token = beepboopaccesstoken

###

DELETE {{baseUrl}}/posts/50
Authorization {{access_token}}
```

The team I work with all use VSC**e and so in the project root there's a `.vscode/requests.http` file with various requests to the services we're building, which is also checked into git. I was hoping to be able to use rest.nvim with this file, but it is filled with file variables like in the example above. I think file variable support would be really cool.

teto:
is that some kind of standard or just this vscode extension's format ?

gegnew:
this would be so nice to have!

teto:
apparently fixed buy #152 feel free to reopen if not the case

mawkler:
That's great! Thank you very much @theNEOkun!

It would be great now if the [Treesitter parser for HTTP](https://github.com/rest-nvim/tree-sitter-http) supported this syntax. I noticed that it lives under the same GitHub organization, and I've created an issue here: rest-nvim/tree-sitter-http/issues/6.

mawkler:
@NTBBloodbath Was this functionality removed in the new release?

NTBBloodbath:
> @NTBBloodbath Was this functionality removed in the new release?

Hi, nope. It still exists, only now if the variable contains text it must be enclosed in a string (`@foo = "https://..."`) :)

mawkler:
@NTBBloodbath I see, thank you!

PR #152, opened 2022-11-16
Variable substitution

theNEOkun:
Added the ability to declare variables by starting them with @, and then using them as environmental variables, Ala [rest-client](https://marketplace.visualstudio.com/items?itemName=humao.rest-client).

I.E:
@address = localhost
@port = 3000
@base = api
@url = http://{{ address }}:{{ port }}/{{ base }}

would result in url = http://localhost:3000/api

// Had to redo everything to fix the history

### DIFF
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -174,7 +174,7 @@ local function end_request(bufnr)
   end
   utils.move_cursor(bufnr, linenumber)
 
-  local next = vim.fn.search("^GET\\|^POST\\|^PUT\\|^PATCH\\|^DELETE", "cn", vim.fn.line("$"))
+  local next = vim.fn.search("^GET\\|^POST\\|^PUT\\|^PATCH\\|^DELETE\\|^//\\|^#", "cn", vim.fn.line("$"))
 
   -- restore cursor position
   utils.move_cursor(bufnr, oldlinenumber)
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -174,7 +174,7 @@ local function end_request(bufnr)
   end
   utils.move_cursor(bufnr, linenumber)
 
-  local next = vim.fn.search("^GET\\|^POST\\|^PUT\\|^PATCH\\|^DELETE\\|^//\\|^#", "cn", vim.fn.line("$"))
+  local next = vim.fn.search("^GET\\|^POST\\|^PUT\\|^PATCH\\|^DELETE", "cn", vim.fn.line("$"))
 
   -- restore cursor position
   utils.move_cursor(bufnr, oldlinenumber)
--- a/lua/rest-nvim/utils/init.lua
+++ b/lua/rest-nvim/utils/init.lua
@@ -48,27 +48,65 @@ end
 -- variables
 M.read_env_file = function()
   local variables = {}
-  local env_file = "/" .. (config.get("env_file") or ".env")
 
-  -- Directories to search for env files
-  local env_file_paths = {
-    -- current working directory
-    vim.fn.getcwd() .. env_file,
-    -- directory of the currently opened file
-    vim.fn.expand("%:p:h") .. env_file,
-  }
+	-- If there is a line at the beginning with @ first
+	if vim.fn.search('^@', 'cn') > 0 then
+		-- Read all lines of the file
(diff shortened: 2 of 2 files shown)


aldevv:
that's sick!

lua/rest-nvim/request/init.lua
```diff
@@ -174,7 +174,7 @@ local function end_request(bufnr)
   end
   utils.move_cursor(bufnr, linenumber)
 
-  local next = vim.fn.search("^GET\\|^POST\\|^PUT\\|^PATCH\\|^DELETE", "cn", vim.fn.line("$"))
+  local next = vim.fn.search("^GET\\|^POST\\|^PUT\\|^PATCH\\|^DELETE\\|^//\\|^#", "cn", vim.fn.line("$"))
```
udayvir-singh, line 177: I don't think parsing blocks based on comments is a good idea, as:
1. it goes against rfc.
2. comments by design shouldn't interact with the code.

lua/rest-nvim/utils/init.lua
```diff
@@ -48,27 +48,65 @@ end
...
+			-- Get all that starts with @
+			if line:match("^@") then
+				-- Read them off
+				local place = M.split(line, '=', 1);
+				-- Remove whitespace around them
+				for i, each in ipairs(place) do
+					place[i] = each:gsub(" ", "");
+				end
+				-- Add to variables, and remove the @
+				variables[place[1]:gsub('@', '')] = place[2];
```
udayvir-singh, line 68: There should be a `nil` check here otherwise it will throw errors.

lua/rest-nvim/utils/init.lua
```diff
@@ -48,27 +48,65 @@ end
...
-  return variables
+	-- If there is a line at the beginning with @ first
+	if vim.fn.search('^@', 'cn') > 0 then
+		-- Read all lines of the file
+		local lines = vim.api.nvim_buf_get_lines(0, 0, -1, true);
+
+		-- For each line
+		for _, line in pairs(lines) do
+			-- Get all that starts with @
+			if line:match("^@") then
```
udayvir-singh, line 60: Use much more concrete regex instead of just `^@`, a much more viable option would be `^@[%w!@#$%^&*-_+?~]+%s*=%s*[^=]+`

lua/rest-nvim/utils/init.lua
```diff
@@ -48,27 +48,65 @@ end
 -- variables
 M.read_env_file = function()
```
udayvir-singh, line 49: Name of `read_env_file` should be changed something like `get_variables` as it no longer just reads env file.

udayvir-singh (changes requested):

lua/rest-nvim/utils/init.lua
theNEOkun: From what I can tell, there does not seem to be any need for a nil-check. There will always be 2 values in 'place'.

If the left side is not present, then it is not evaluated, if the right side is not present then a "" is placed there. string:gsub as well only removes if it exists.

udayvir-singh: yeah, this comment was made before there was `^@` regex in place.

lua/rest-nvim/utils/init.lua
```diff
@@ -43,32 +43,70 @@ M.read_file = function(file)
...
-  return variables
+	-- If there is a line at the beginning with @ first
+	if vim.fn.search('^@', 'cn') > 0 then
+		-- Read all lines of the file
+		local lines = vim.api.nvim_buf_get_lines(0, 0, -1, true);
+
+		-- For each line
+		for _, line in pairs(lines) do
+			-- Get all that starts with @
+			if line:match("^@[%w!@#$%^&*-_+?~]+%s*=%s*[^=]+") then
```
udayvir-singh, line 60: Since we are already using a string.match here. Code from line 60-67 can just be compressed into a single line using capture groups in lua regex.

```lua
local name, val = line:match("^@([%w!@#$%^&*-_+?~]+)%s*=%s*([^=]+)$")

if name then
  variables[name] = val
end
```

udayvir-singh (changes requested):

udayvir-singh (approved):

udayvir-singh:
closes #132

NTBBloodbath (approved):
Hey,

That's sick, thank you so much for doing this!

Also thanks to @udayvir-singh for the suggested changes and code review, everything looks awesome :)

PR #159, opened 2022-12-12
feat: introduce a testing framework

teto:
Run 'make test' to run tests.

Tests could drive the API towards a more programmatic interface.
This is just a bootstrap to future work.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -1,0 +1,14 @@
+name: default
+
+on: [push, pull_request]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v18
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+    - run: |
+        nix develop ./contrib -c make test
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,5 +1,5 @@
 -- Use lua52 so we will no receive errors regarding to goto statements
-std = 'lua52'
+std = 'lua52+busted'
 
 -- Rerun tests only if their modification time changed
 cache = true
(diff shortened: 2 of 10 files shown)


teto:
I wish we can improve the API for programmatic usage. Right now, output is intertwined with the code.
When testing rest.nvim there are 2 distinct aspects:
1/ testing the actual request (which sadly requires internet, I would like it better if we could run the test locally)
2/ testing the parsing of http files, which could really use a better API: for instance we could check that the number of requests in a file is the correct one. Right now we can just iterate request by request.

teto:
1/ is low priority for me. I had a look at something like reqres.in that we could run locally and found https://github.com/mockoon/mockoon/tree/main/packages/cli#installation. I wanted something like this to write a nixosTest later on (in which you dont have internet access) or to test when you are offgrid (in train for instance) but reqres.in is good engouh.

As for 2, I will explore a bit of fennel and a bit of teal to satisfy my own curiosty, so that may give me some ideas. I think we can iteratively improve the current api, it's not like there are consumers of it yet. We can make it purer (aka avoid sideeffects like moving cursor or printing to screen) bit by bit.

For instance getting the number of total requests

One thing I wonder though is about the interoperability with other tools. I dont think we have too: we need to leverage lua and neovim integration as is done with a custom syntax in https://github.com/rest-nvim/rest.nvim/pull/122, but maybe we can make it less alien or easier to import other formats to/from. I will be playing with https://hurl.dev/ a bit I think

teto:
I didn't want to impose nix to anyone but this is just the most reliable way to have reproducible builds/pipelines. I've added some doc in the README how to run the tests. If someone wants to sink in time to make it work without nix, he is welcome but that wont be me (ideally keeping compat with both). There is redundancy in the flake but I will fix that later, ideally I would like to merge this first so that I can iterate on some minimal testing and merge https://github.com/rest-nvim/rest.nvim/pull/122

Issue #157, opened 2022-12-02
use `content-type:`, other casing doesn't work


xbladesub:
I used rest.nvim couple of weeks ago and everything was fine. However it doesn't work anymore and I have no idea why. 

This is my request:
```
POST http://127.0.0.1:8080/v1/user/signup
Content-Type: application/json
API-TOKEN: API-TOKEN: 0a479179624fb5255cb9054603382077dd7366b0319386dd89b12806463bd000

{
  "device_id": "admin_1",
  "password": "admin_1"
}
```

I use Rust - actix-web framework. It says ```Error in response: Deserialize(Error("expected value", line: 1, column: 1))```

Btw, everything is okay in Insomnia and Postman

jungai:
Same issue, maybe because of this

```bash
curl -sSL --compressed -X 'POST' -H 'Content-Type:  application/json' -d 'key=value' -d 'key=value' -d 'key=value' 'url'
```

Emptyfruit:
Can be related to https://github.com/rest-nvim/rest.nvim/issues/114 because I got this issue after installing today for the first time.

sjdonado:
Same here. This project seems to be a bit abandoned :(

rhinoxi:
I think this is related to the PR https://github.com/rest-nvim/rest.nvim/pull/142 in this thread https://github.com/rest-nvim/rest.nvim/issues/141#issuecomment-1330348198, which makes "Content-Type: application/json" couldn't be recognized and the brace around the raw body will be removed when calling curl.

MikaelElkiaer:
> Same here. This project seems to be a bit abandoned :(

It does seem abandoned.

If you see the branch [document-competition](https://github.com/rest-nvim/rest.nvim/tree/document-competition#related-software) there is an alternative [Hurl](https://github.com/Orange-OpenSource/hurl) that seems a lot more mature, although it is not specifically for Neovim.
There is also a [tree-sitter grammar](https://github.com/otherJL0/tree-sitter-hurl) and you could probably use something like [flow.nvim](https://github.com/arjunmahishi/flow.nvim) to execute - combining the two and you could do a visual via tree-sitter and run with flow.

sharksforarms:
for those looking for a quick work-around, replace `Content-Type` with `content-type`

teto:
[...]

PR #162, opened 2022-12-27
test: programmatic run of .http file

teto:
We want to be able to run a whole .http file . 

The dance around cursor position is quite annoying.


teto:
I tested `:RestNvim` and it seemed to work, minus this issue which is next https://github.com/rest-nvim/rest.nvim/issues/157

PR #165, opened 2022-12-29
fix formatter for html and error in rest.run()

### LINKED ISSUE #164: Disable debug logs?
Is it possible to disable debug logs, some debug logs are not very helpful for user.
e.g.
```
[rest.nvim] [DEBUG 14:43:41] ...pack/packer/opt/rest.nvim/lua/rest-nvim/request/init.lua:179: Searching pattern starting from 23
```

udayvir-singh:
Fixes error caused by html formatter running even when tidy is not available on the system. Also fixes docs related to disabling formatters.

fixes #164

### DIFF
--- a/lua/rest-nvim/init.lua
+++ b/lua/rest-nvim/init.lua
@@ -21,7 +21,7 @@ rest.run = function(verbose)
     return
   end
 
-  return rest.run_request(result, verbose)
+  return rest.run_request(result, {verbose = verbose})
 end
 
 -- run will retrieve the required request information from the current buffer
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ use {
         show_http_info = true,
         show_headers = true,
         -- executables or functions for formatting response body [optional]
-        -- set them to nil if you want to disable them
+        -- set them to false if you want to disable them
         formatters = {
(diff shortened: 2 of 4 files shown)


udayvir-singh:
Now this PR also fixes #164 and fixes nasty error on rest.run() due to incorrect handling of opts.

lua/rest-nvim/init.lua
```diff
@@ -21,7 +21,7 @@ rest.run = function(verbose)
     return
   end
 
-  return rest.run_request(result, verbose)
+  return rest.run_request(result, {verbose = verbose})
```
ray-x, line 24: this will fix a breaking change.

ray-x (commented):
Thanks. the changes fix https://github.com/rest-nvim/rest.nvim/issues/164

ray-x (approved):
LGTM

teto:
you can drop the second commit. I think it's best to throw an error if tidy is not available, so nothing to change on that side, what I am curious about is why `html = nil` didn't work (that's the way to remove entries from lua arrays). I believe it might be due to ` config = vim.
tbl_deep_extend("force", config, user_configs)`

PR #169, opened 2022-12-29
chore: dont special case json

teto:
we special case json when checking the payload body via
encoding/decoding it.
This is done by checking the "content-type" in:
https://github.com/rest-nvim/rest.nvim/issues/157

There doesn't seem to be a need for that other than possibly validate
the payload ? in which case we should implement something more generic.

### DIFF
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -41,8 +41,7 @@ end
 -- @param bufnr Buffer number, a.k.a id
 -- @param start_line Line where body starts
 -- @param stop_line Line where body stops
--- @param has_json True if content-type is set to json
-local function get_body(bufnr, start_line, stop_line, has_json)
+local function get_body(bufnr, start_line, stop_line)
   -- first check if the body should be imported from an external file
   local importfile = get_importfile_name(bufnr, start_line, stop_line)
   local lines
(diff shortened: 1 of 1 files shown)


teto:
@udayvir-singh  what do you think ?

udayvir-singh:
@teto fixed with #170

lua/rest-nvim/request/init.lua
```diff
@@ -67,23 +66,6 @@ local function get_body(bufnr, start_line, stop_line, has_json)
...
-      -- convert nested tables to string.
-      for key, val in pairs(json_body) do
-        if type(val) == "table" then
-          json_body[key] = vim.fn.json_encode(val)
-        end
-      end
-      return json_body
-    end
-  end
-
```
udayvir-singh, line 86: DONOT delete these lines; this statement does two things:
1. LINE 75: Minify json. ie. remove extra tabs and spaces to reduce bandwidth.
2. LINE 78-83: Parse nested json into valid `application/x-www-form-urlencoded` string values. see #138

lua/rest-nvim/request/init.lua
```diff
@@ -67,23 +66,6 @@ local function get_body(bufnr, start_line, stop_line, has_json)
     end
   end
 
-  local is_json, json_body = pcall(vim.fn.json_decode, body)
-
-  if is_json then
```
udayvir-singh, line 72: Don't delete these lines; this statement does two things:
1. LINE 75: Minify json. ie. remove extra tabs and spaces to reduce bandwidth.
2. LINE 78-83: Parse nested json into valid `application/x-www-form-urlencoded` string values. see #138

teto: I think 1 is non-relevant, also it means we dont send the exact same payload as specified by the user which is concerning (encoding/decoding floats can raise issues I suppose, same for numbers ?).

As for 2, application/x-www-form-urlencoded seems like a pain https://stackoverflow.com/questions/9870523/what-are-the-differences-between-application-json-and-application-x-www-form-url but indeed that's the curl default. One solution could be to default to application/json but that can be surprising too.

Maybe we could do is implement the same mechanism as for "formatter", but for inputs instead of outputs. have the ability to to preprocess the payload depending on content-type.

PR #170, opened 2022-12-30
fix: formatter for html and special casing of content-type

### LINKED ISSUE #157: use `content-type:`, other casing doesn't work

udayvir-singh:
## FIX 1
Fixes error caused by html formatter running even when tidy is not available on the system. Also fixes docs related to disabling formatters.

> This PR is a cleaner version of #165.

---

**Reason why formatter should be set to `false` instead of `nil`:**

`nil` means nothing, ie. absence of value; that's why you have how pass `false` instead which has a actual value `0` in binary which can be set to a variable and passed around.

Here's another analogy:

Let's say you asked me to put something in a box, and I put `<nothing>` in the box. Is the box now empty or does it contain `nothing`? There is no way for you to verify that.

## FIX 2

Fixes #157  hardcoded lower casing of content-type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ use {
         show_http_info = true,
         show_headers = true,
         -- executables or functions for formatting response body [optional]
-        -- set them to nil if you want to disable them
+        -- set them to false if you want to disable them
         formatters = {
           json = "jq",
           html = function(body)
--- a/lua/rest-nvim/config/init.lua
+++ b/lua/rest-nvim/config/init.lua
@@ -16,6 +16,9 @@ local config = {
     formatters = {
       json = "jq",
       html = function(body)
+        if vim.fn.executable("tidy") == 0 then
+          return body
+        end
         -- stylua: ignore
         return vim.fn.system({
           "tidy", "-i", "-q",
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -67,7 +67,7 @@ local function get_body(bufnr, start_line, stop_line, has_json)
     end
   end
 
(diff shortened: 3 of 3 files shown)


teto:
> nil means nothing, ie. absence of value; that's why you have how pass false instead which has a actual value 0 in binary which can be set to a variable and passed around.

setting the value to nil is the standard way of removing a key from a table http://lua-users.org/wiki/StoringNilsInTables so I was surprised by the current behavior, which may be an oddity of tbl_deep_extend (maybe worth documenting upstream). I will test then merge

udayvir-singh:
@teto

From lua-users wiki:
> Lua tables make no distinction between a table value being nil and the corresponding key not existing in the table

> In fact, you may think of {} as a table with all possible keys set to nil, and this still takes only a small finite amount of memory because all those keys having nil values are **not explicitly stored.**
 
So basically coming back to my original comment, the following two tables are indistinguishable:
```lua
local tbl1 = {}

local tbl2 = {
  a = nil,
  b = nil,
  c = nil,
  d = nil,
  ... -- all the way to infinity
}
```

PR #174, opened 2023-01-06
try to use treesitter instead

teto:
trying to use treesitter nodes to identify the requests, method etc instead of relying on weak heuristics.

The issue I have at the moment is the root node appears as a "document" https://github.com/rest-nvim/tree-sitter-http/blob/2c6c44574031263326cb1e51658bbc0c084326e7/grammar.js#L15 but I can't find it anywhere in the treesitter playground

TODO:
- [ ] GET remove or ditch support for this syntax ?
```
/api/users?page=5 HTTP/1.1
Host: reqres.in:443
```
- [ ] support of curl_raw_args ?
- [ ] parse_url had `if config.get("encode_url") then  target_url = utils.encode_url(target_url)  end`, do we still need this ?
- [ ] check one can set http_version as in older version
- [ ] remove regex to capture the "script" part and instead retreive the content of the "internal_payload" node (problem is that currently the "internal_payload" node is full of errors)
- [x] support external body
- [ ] support embedded body

we already have variables
- [x] should be able to list requests in a file

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,3 +1,4 @@
+-- List of warnings at https://luacheck.readthedocs.io/en/stable/warnings.html
 -- Use lua52 so we will no receive errors regarding to goto statements
 std = 'lua52+busted'
 
@@ -6,6 +7,8 @@ cache = true
 
 ignore = {
 	'631', -- max_line_length
+    '213',
+    '214'
 }
 
 read_globals = {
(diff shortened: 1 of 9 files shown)


NTBBloodbath:
This was already done after the parser refactor so it makes no sense to have this draft PR stalled :p

PR #175, opened 2023-01-14
fix: body should be present only in supported methods

vodenkaj:
Added condition on whether or not to include `--data-raw` parameter to the command. Only `POST`, `PUT`, `PATCH`, and `DELETE` should contain such parameters. 

### GET request with headers:
20230114_15h37m35s_grim
**CURRENT**
Result: 
`curl -sSL --compressed -X 'GET' -H 'Authorization: Bearer Test.Token' --data-raw 'authorization: Bearer Test.Token' 'https://test.com/auth/token/list'`
**EXPECTED**
Result: 
`curl -sSL --compressed -X 'GET' -H 'Authorization: Bearer Test.Token' 'https://test.com/auth/token/list'`

This one is critical since requests to the google cloud APIs, for instance, will reject such `GET` request.
20230114_15h49m26s_grim

[HTTP Methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods)

### POST request without a body
20230114_15h43m57s_grim
**CURRENT**
Result: 
`curl -sSL --compressed -X 'POST' -H 'Authorization: Bearer Test.Token' --data-raw 'authorization: Bearer Test.Token' 'https://test.com/auth/token/list'`
**EXPECTED**
Result: 
`curl -sSL --compressed -X 'POST' -H 'Authorization: Bearer Test.Token' --data-raw '' 'https://test.com/auth/token/list'`

Since the `vim.api.nvim_buf_get_lines` function accepts a zero-based index, and we want to ignore the line between headers and body, decreasing the value by 1 is incorrect.

### DIFF
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -52,7 +52,7 @@ local function get_body(bufnr, start_line, stop_line, has_json)
     end
     lines = utils.read_file(importfile)
   else
-    lines = vim.api.nvim_buf_get_lines(bufnr, start_line - 1, stop_line, false)
+    lines = vim.api.nvim_buf_get_lines(bufnr, start_line, stop_line, false)
   end
 
   local body = ""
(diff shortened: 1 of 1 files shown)


teto:
you can perfectly send a payload for get requests

Issue #180, opened 2023-01-23
dont automatically rewrite the .env file


teto:
In https://github.com/rest-nvim/rest.nvim/pull/122 we merged support for env session 

the current behavior is that on any variable change, the .env file is regenerated which I find very annoying. 
Rewriting the file discards commits an the structure, plus in case of failure it can destroy the file. I tend to think that .env are often not (git) versioned due to their dynamic nature so losing that file can be surprising as well as undesired. 

As I develop on rest.nvim, this happens quite often to me. 
The ability to save a run to a file is enticing but I think we should let the user be the master of that in a 
```
{% 
context.write_to_file ...
%}
```

One problem too is that set_env only accepts strings so far, as the variable will be serialized at some point, while for my tests, I would like to add any type of variable to the context.

The context could be buffer-local and shared between requests of a same buffer. 
cc @D-James-GH

D-James-GH:
A request from my own workflow would be that the set variable would work across buffers and is accessible until neovim is closed. 
My main use for the set_env is to save my access token from the response of "/auth". I then use this in the header of all other requests. 
For my workflow I would also need the value of whatever I set to be available when rest.nvim is inputting variables so that the following works: 
```
GET {{base_url}}/posts
Authorization: Bearer {{access_token}}
```
Raw writing to a file is available already as part of the global lua environment. For example, the following creates and writes "hello world" to a file: 
```
GET https://jsonplaceholder.typicode.com/posts/1
Content-type: application/json

{%
local file = io.open("test.txt", "w")
file:write("Hello World")
file:close()
%}

```  
So I don't know if it is necessary to add an extra field to the context table with this ability. 

However, as you mentioned in #122, it would be very good to have the option to choose whether "set_env" persists to the currently selected env file or just to memory. 

You also mentioned, in #122, an interface like: 
```
{% 
context.set_env("some_key", "value",  "write_flag (defaults to false)")
%}
```
This solution would work well for my work flow. And if we changed "value" to be any type not just string then it would cover this issue as well.

How would something like that work for your workflow?

teto:
or just provide different variable context: persistent-context (with an autosave to file option to mimic what is available), or a buffer-based temporary context (my use case is testing a CRUD scenario where I post -> get -> put -> get -> delete but dont care about the resources, I just need to retreive the generated IDs on post/put)

gautaz:
[...]

PR #181, opened 2023-02-18
Feature/variable management

D-James-GH:
This pull request satisfies #180. 

It adds two ways of dealing with variables from within the script tag.

1. Global environment variables

Same syntax as before, however, with the option to persist the data to the currently selected Env file. If `persist` is `false` (the default) then the variable is available until the editor is closed. If `persist` is `true` then the variable is updated/written to the env file, this logic is the same as the current logic.
```lua
{% 

-- written to file
context.set_env("postTitle", context.json_encode(body.title), { persist=true })

-- saved in session memory
context.set_env("userId", body.userId)

%}
```

3. Buffer variables
Buffer variables are only available in the current buffer, and implemented using `vim.api.nvim_buf_get_var`
```lua
{% 

context.set_buffer_variable("postId", body.id)

%}
```


teto:
you should have pinged me I just see it. will have a look later

lua/rest-nvim/variables/init.lua
teto, line 1: I feel "variables" is a bit generic as a module name (no strong opinion though). What about "environment" or "context" ? I dont know about postman/hurl conventions , hurl calls them template  https://hurl.dev/docs/templates.html

D-James-GH: yeah, fair point. "environment" makes sense to me as it handles the runtime environment

teto:
you've updated the tests and all, that's great.
I also wonder if instead of having dedicated functions like `context.set_buffer_variable(`  we shouldn't have several contexts:
bcontext (buffer) and gcontext (global) so that both can have the same interface aka bcontext.set_variable and gcontext.set_variable (gcontext could be just context, also it's a bit verbose so we could even shorten it to `ctx` maybe).
Another possibility would be to use `context.set_variable(..., { buffer = true})` but it is more verbose.

What do you think ?

D-James-GH:
> Another possibility would be to use `context.set_variable(..., { buffer = true})` but it is more verbose.

I personally prefer this as I think it's easier to understand what it does. It also looks like many other neovim functions in use. The only thing that might bit confusing is the writing to env file. If `set_env` is combined with `set_variable` then `buffer = true` wouldn't do anything if a `set_env = true` flag is set because set_env would always be global. Here's what I am thinking in code:
```lua
-- Sets a variable for the current buffer only.
context.set_variable("key", value.item, { buffer = true})

-- Sets a global runtime variable.
context.set_variable("key", value.item, { buffer = false})

-- Writes to the currently selected env file. Here the "buffer" flag does nothing.
context.set_variable("key", value.item, { buffer = false, set_env_file = true})
```

boltlessengineer:
Closing this due to v3 release.
rest.nvim will follow the api of intellij’s http client.

Now rest.nvim provides three types of variables:

- environment variables (variables from your current shell session)
- in-place variables (variables defined in file with @name=value syntax)
- client/request variables (variables defined from request/response handler scripts)

PR #187, opened 2023-03-09
fix: include documentation on rockspec

Leixb:
Fixes the missing documentation was missing when installing through luarocks.

Also, I noticed the following error when checking the `luarocks` build:

```
Error: Inconsistency between rockspec filename (rest.nvim-scm-1.rockspec) and its contents (rest.nvim-0.1-2.rockspec).
```

I changed both to `scm-3` but maybe another major is better.

### DIFF
--- a/rest.nvim-scm-1.rockspec
+++ b/rest.nvim-scm-1.rockspec
@@ -32,6 +32,7 @@ end
 build = {
    type = "builtin",
    copy_directories = {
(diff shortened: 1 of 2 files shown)


teto:
The luarocks tag release should generate better rockspecs, just we haven't cut a release with luarocks-tag-release improvements. This is still useful for the scm version indeed

PR #190, opened 2023-04-12
fix: find next request line when the verb is 'DELETE'

Jose-Luis:


NTBBloodbath (approved):
Thank you so much for this!

Issue #89, opened 2021-12-14
pass credentials for http basic auth


joshuali925:
Using curl, credentials can be passed with the `-u <user>:<pass>` flag:

```
curl -u guest:guest https://jigsaw.w3.org/HTTP/Basic/
```

But I didn't find the corresponding option here, how to do the same in rest.nvim? thanks

---

I took another look and seems auth is not supported yet? A simple fix would be to also exclude `@` when encoding url
https://github.com/NTBBloodbath/rest.nvim/blob/3ab950fe88cf836c83fa15a10e386ccd5f6aa1cb/lua/rest-nvim/utils/init.lua#L273-L274

According to https://stackoverflow.com/a/19737890 `@` can be used directly in url, and then this will work

```
GET https://guest:guest@jigsaw.w3.org/HTTP/Basic/
```

NTBBloodbath:
Hi, thank you so much for the information. Being honest I found this too a few days ago when trying out an API.

I have read plenary's curl wrapper source again and it seems that they have auth as an undocummented feature. I'll start doing an implementation in our side to interact with that "hidden" feature but also include the `@` in url so user will have both options :)

PR #115, opened 2022-05-16
Interpret "Authorization: Basic <user>:<pass>" header as basic auth

srihari93:
Trying to add the ability to interpret a special header line `Authorization: Basic <user>:<pass>` as basic authorization. 
This follows the VS code rest plugin's[ style of interpreting basic auth](https://github.com/Huachao/vscode-restclient/blob/master/README.md?plain=1#L307-L309).

---------
```rest
POST https://fake.com/api/some
Authorization: Basic user:password
Content-Type: application/json

{
  "key": "jafa"
}
```
is converted to `curl -sSL --compressed -X 'POST' -H 'Content-Type:  application/json' -d 'key=jafa' -u 'user:password' 'https://fake.com/api/some'`

----------

@NTBBloodbath 
There is an alternative suggestion [here](https://github.com/NTBBloodbath/rest.nvim/issues/89) on how to do the basic auth, both of these styles can live together, IMHO. I went for this style as its compatible with the heavily used VS code. Feel free to reject this if the VS code style is not palatable. Also fine if you have issues with style, let me know I will fix and get back to you.

PR #192, opened 2023-04-26
Exclude encoding `@` to support `user:pass@host` authentication

joshuali925:
fixes https://github.com/rest-nvim/rest.nvim/issues/89

I see there's https://github.com/rest-nvim/rest.nvim/pull/115, but this is a one-character change, would you like to consider it? Thanks anyways


NTBBloodbath (approved):

NTBBloodbath:
Thank you for this small patch!

PR #197, opened 2023-05-31
build(nix): move flake to root + add busted to luad env

teto:
I could do without but I prefer to have the flake at the root, I just have to use `nix develop` instead of `nix develop ./contrib`. It's also more discoverable and because we dont have that many files at the root, I feel like it's ok.
@NTBBloodbath do I have your blessing ?


NTBBloodbath:
Yeah I don't see any issues with moving it to root, feel free to merge this if you want to. Also don't you need to also move the flake lockfile or something like that?

teto:
yes but I moved the lockfile as well already so all is fine

PR #200, opened 2023-06-06
dont display binary outputs in answer

teto:
I was testing an API that returned a zip file and it was getting out of
hand so I added an is_binary_content_type to control whether to display
the content type.
Ideally I would have liked the display to be a default postprocess step so
that it gets more generic and let the user more control but this needs
deeper refactoring.


teto:
I guess we should sitll display it in case of error, aka still inline error if  status is not ok

PR #207, opened 2023-06-11
feat: show pinner while runnig request

antosha417:
Hey! I really enjoy using this plugin, use it almost every day. 
Some queries that I run takes long time to complete. And while they are running I don't get any feedback from neovim. 
Decided to add a little spinner on the line of the request so I can know that the request is running 

Took an inspiration of this in https://github.com/hudclark/grpc-nvim . It is similar plugin but for grpc calls

### DIFF
--- a/lua/rest-nvim/curl/init.lua
+++ b/lua/rest-nvim/curl/init.lua
@@ -1,4 +1,5 @@
 local utils = require("rest-nvim.utils")
+local spinner = require("rest-nvim.utils.spinner")
 local curl = require("plenary.curl")
 local log = require("plenary.log").new({ plugin = "rest.nvim" })
 local config = require("rest-nvim.config")
@@ -68,10 +69,11 @@ M.get_or_create_buf = function()
   return new_bufnr
 end
 
-local function create_callback(curl_cmd, method, url, script_str)
+local function create_callback(curl_cmd, method, url, script_str, stop_spinner)
   return function(res)
     if res.exit ~= 0 then
       log.error("[rest.nvim] " .. utils.curl_error(res.exit))
+      stop_spinner(true)
       return
     end
     local res_bufnr = M.get_or_create_buf()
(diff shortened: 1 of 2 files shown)


teto:
I confess I am not a fan of the feature, I dont have a use of it and also the spinner looks like a generic thing you would want for several plugins. Something I've wanted for a while is to provide hooks for users to plug some stuff in. In this case we could fire an event when launching a request and when finishing it. This way you could subscribe to both events to start/stop the spinner. What do you think ?

teto:
to be clear, I dont have time to implement this but as @NTBBloodbath seems to agree with the idea, I think we could merge a PR implementing that.

Quickly how it could work:
- fire a User RestNvimStartRequest <ID> event upon `run_request` ( `:h User`, the ID could be some hash of the request or an autoincrementing id)
- fire a User RestNvimStopRequest .

If we want to support cancel events, we could do `RestNvimRequest start/stop/cancel id` instead to avoid using too many event names.
I think such a callback could be useful to time requests too (though that may be less precise than the time reported by curl)

teto:
I was going through the PR list and several PRs could use the previous mechanism. I've implemented a quick draft
 https://github.com/rest-nvim/rest.nvim/pull/214 that can serve as inspiration.
 @antosha417 would you be willing to test/amend/fork that PR so that you can achieve what you initially wanted with a callback based approach ? Feel free to open a new PR and I can review/merge it

NTBBloodbath:
Note for myself: perhaps this can be implemented in future versions using [fidget.nvim](https://github.com/j-hui/fidget.nvim) through luarocks.

boltlessengineer:
Closing this PR due to the v3 release. rest.nvim now supports spinner using [fidget.nvim](https://github.com/j-hui/fidget.nvim)

antosha417:
@boltlessengineer that is a great news! Thank you

PR #210, opened 2023-06-27
feat: add search_back option

jetgeng:
In most instances, users cursor is below start_request line. if user want to  send request, he/she must jump to one line which above or the start_request line. so i add a option, allow user change search direct.


teto:
could you share an example ? I think this might be a bug. Regardless the treesitter branch should allow for what you are asking for but it needs more work.

jetgeng:
i clip some screen as follow:
image
image
image
I dig the code, found the code at  [request/init.lua](https://github.com/rest-nvim/rest.nvim/blob/f1597ab6df09b0f04196fa7d516a1b20c91f292e/lua/rest-nvim/request/init.lua#L221). 
just search below current line
```lua
local res = vim.fn.search("^GET\\|^POST\\|^PUT\\|^PATCH\\|^DELETE", "cn")
```
if  search back, change "cn" to "cb". we can get affect like below:
image

PR #214, opened 2023-07-11
add callbacks on start/stop request

teto:
This is a quick & dirty implementation of https://github.com/rest-nvim/rest.nvim/pull/214, e.g., rest-nvim throws events when 
- launching a request
- finishing a request

This is what I intented to test it with
```
local start_time = 0

local on_start_request =  function (req)
  -- vim.loop.gettimeofday()
  -- vim.fn.reltime()
 start_time = vim.loop.now()
 print("starting request")
end

local on_stop_request =  function  (...)
 print("finished request")
 -- TODO displayu time taken
end

vim.api.nvim_create_autocmd("User", {
pattern = "RestStartRequest",
once = true,
  callback = function()
    print("IT STARTED")
  end,
})
```


teto:
this might not be perfect but there is another PR I wanna submit and I got limited time. I tested this quickly and it looked ok, we can fix it later if we wanna pass more information in callbacks

PR #216, opened 2023-07-11
dont inline file

teto:
- doc: mention debug instructions
- feat: make it possible to not inline external file

if you reference a file via <@ file instead of < file the filename will be passed as is to curl, thus making the curl command in preview a ton shorter.


teto:
I ran a few tests looks ok. It's a gamechanger for big payloads, for instance with a big json, the request goes from a minute to instant.

PR #217, opened 2023-07-11
separate parsing from running steps

teto:
the treesitter backend is promising but it is too much work to reach feature parity with our current parser: 
https://github.com/rest-nvim/rest.nvim/pull/174

what I am proposing here is a first refactoring to delay variable splicing until last minute. 
It untangles parsing & running requests. Right now you need variables to exist when parsing which means you have to execute the request when parsing them else you get errors.
This PR replaces (aka "splices") the variables at the last minute so you can parse requests without executing them.
It allowed me to add a `buf_list_request` function in the current backend, which was not possible before, so now `run_file` is backend agnostic (it first lists requests and then run them).

In this PR, you can't change backend yet but in a future one, I will add a treesitter backend to parse request.
We will then be able to switch  (optional) to a treesitter backend even though it's not complete (One could even imagine a hurl backend) which will make working on it easier.

### DIFF
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -1,5 +1,4 @@
 local utils = require("rest-nvim.utils")
-local path = require("plenary.path")
 local log = require("plenary.log").new({ plugin = "rest.nvim" })
 local config = require("rest-nvim.config")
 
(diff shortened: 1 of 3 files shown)


lua/rest-nvim/request/init.lua
```diff
@@ -41,55 +33,34 @@ end
...
+  local lines -- an array of strings
   if importfile ~= nil then
-    if not utils.file_exists(importfile) then
-      error("import file " .. importfile .. " not found")
-    end
-    lines = utils.read_file(importfile)
+    return { external = true, filename_tpl = importfile }
   else
-    lines = vim.api.nvim_buf_get_lines(bufnr, start_line - 1, stop_line, false)
+    lines = vim.api.nvim_buf_get_lines(bufnr, start_line, stop_line, false)
```
teto, line 44: I am a bit afraid removing the -1 might create bugs down the line but it was incorrect, i.e, if you look at the main branch today the request in tests/basic_get.http
```
GET https://reqres.in/api/users?page=5
```
loads with `GET https://reqres.in/api/users?page=5` as body xD

This means we could use golden (also called "snapshot") tests to make sure this is not the case. This PR helps making rest.nvim more "api-like" so I think it's a move in the good direction overall.

teto:
I noticed that we dont tests variable splicing in tests but I've done some manual testing and it looked fine. Planning to add them to tests in next PRs. I think I'll merge and if bugs are spotted, I will correct them ^^''

Issue #215, opened 2023-07-11
add a :RestLog command ?


teto:
Like `:LspLog`, to display the log immediately and help debug the tool

PR #218, opened 2023-07-11
feat: adding a RestLog command

teto:
Fixes https://github.com/rest-nvim/rest.nvim/issues/215


NTBBloodbath:
Looks good to me, as a note I think that we must convert that file to Lua so we can use the Lua APIs. It could be another PR tho

NTBBloodbath (approved):

teto:
> we must convert that file to Lua so we can use the Lua APIs

I thought so too but was afraid to break things with the cpo code.

PR #223, opened 2023-08-13
Fix RestStopRequest event 

ddomurad:
Hello :)

Events implementation as proposed in [214](https://github.com/rest-nvim/rest.nvim/pull/214) seems to work incorrectly. 

Issue 1:
Because a `callback` function is passed to plenary's curl, the request will be made asynchronously, causing the `RestStopRequest` event to be emitted right after the `local success_req, req_err = pcall(curl.curl_cmd, Opts)` returns, way before the actual request finishes. 

Issue 2:
The events are emitted only during normal `run_request`, they are not emitted during rerunning the `last` request. 

Proposed changes: 
1. Move the emition of events into the `curl/init.lua` file. This way it won't matter if the call is made via `run_request` or `last` method.  I'm actually not sure about his one though. Maybe it would be better to extract the events back to the main `init.lua` file, however it was tempting how easy it was to implement. 
2. Emit the `RestStopRequest` inside the curl callback function. 
3. Add a `on_error` handler for plenary's curl. This allows us to emit the `RestStopRequest` event, when the request fails (.e.g. could not resolve the URL)  

The proposed changes will alter: 
1. The payload of `RestStartRequest`  and `RestStopRequest` - this can be some what mitigated I guess. However might be not a big deal, if the feature didn't worked correctly in the first place. 
2. The call stack of the emited error while calling `error(err.message)` in `create_error_handler`

Ofc.  I'm open to any suggestions :) 
I didn't had the time to check if this is in line with other PR that are currently open :(

### DIFF
--- a/doc/tags
+++ b/doc/tags
@@ -8,6 +8,7 @@ rest-nvim-license	rest-nvim.txt	/*rest-nvim-license*
 rest-nvim-quick-start	rest-nvim.txt	/*rest-nvim-quick-start*
 rest-nvim-response-script	rest-nvim.txt	/*rest-nvim-response-script*
 rest-nvim-usage	rest-nvim.txt	/*rest-nvim-usage*
+rest-nvim-usage-callbacks	rest-nvim.txt	/*rest-nvim-usage-callbacks*
 rest-nvim-usage-commands	rest-nvim.txt	/*rest-nvim-usage-commands*
 rest-nvim-usage-dynamic-variables	rest-nvim.txt	/*rest-nvim-usage-dynamic-variables*
 rest-nvim-usage-environment-variables	rest-nvim.txt	/*rest-nvim-usage-environment-variables*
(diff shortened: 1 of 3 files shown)


teto:
thanks for looking into it, I might have been a bit cow-boy about the merge (we need more tests :) ). I will have a look in upcoming days when I am able to use rest.nvim again.

teto:
That looks much better than my initial proposition. I haven't really tested. Do you think you could plug some test ? if not do you think you could do it later once the test infra improves ?

teto (approved):

ddomurad:
I will try to find some time this weekend.

ddomurad:
@teto Well, looking at the current tests implementation, I don't think I will be able to add proper testing for my changes. However I will be glad to help when the proper test infrastructure is there.

teto:
cool. I haven't had time to work on rest.nvim recently, if you are interested in improving tests, you can have a look/resume https://github.com/rest-nvim/rest.nvim/pull/225 , basically it tries to use busted using nvim as a a lua interpreter (via `nvim -l`)

PR #225, opened 2023-08-28
test: use neorocks

teto:
This allows to use a neovim-aware lua interpreter
https://github.com/nvim-neorocks/neorocks/

`$ busted tests/` fails with
```

suite tests/main_spec.lua
...uajit-2.1.0-2022-10-04-env/share/lua/5.1/plenary/log.lua:9: attempt to index global 'vim' (a nil value)

stack traceback:
	...uajit-2.1.0-2022-10-04-env/share/lua/5.1/plenary/log.lua:9: in main chunk
```
but because it ignores the interpreter set in luarocks, which is why the command to run really should be `luarocks test` but in my case it tries to install the busted package and fails: the luarocks config should be updated to contain busted (via nixpkgs' generateLuarocksConfig for instance).

### DIFF
--- a/.busted
+++ b/.busted
@@ -1,0 +1,12 @@
+return {
+  _all = {
+    coverage = false,
+    lpath = "lua/?.lua;lua/?/init.lua",
+  },
+  default = {
+    verbose = true
+  },
+  tests = {
+    verbose = true
+  },
+}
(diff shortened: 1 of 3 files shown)


boltlessengineer:
Closing due to #486

PR #230, opened 2023-09-15
Fix request hightlight

aPisC:
run_request and run_file request opts did not use the highlight.enabled setting from config, the default requestOpts prevented the current request to be hignlighted. Added the missing config to the creation of final request ops


NTBBloodbath (approved):
Woops, nice catch!

Thank you so much for the patch :)

PR #236, opened 2023-09-26
Add case-insensitive table lookups, and use for host and content-type.

### LINKED ISSUE #163: Host or host?
In sample:
```rest
GET /api/users?page=5 HTTP/1.1
Host: reqres.in:443
```
But in code it check `host`:

https://github.com/rest-nvim/rest.nvim/blob/c5094a8ae90e3d2411f4de988c379c863593f0e0/lua/rest-nvim/request/init.lua#L276-L281

This makes the sample not work,
Is the sample incorrect or the request/init.lua is incorrect?

PhilRunninger:
Closes #163.

1. A new utility function called `key` takes a table and a key. It loops through the table items, comparing the each key with the given search key in a case-insensitive manner. Return the first matching key; otherwise, return the given search key. This allows operations like these to be performed:

    ```lua
    local t = {s=37, S=73}
    t[utils.key(t,'a')] = 42   -- Insert a=42             t == {s=37,S=73,a=42}
    y = t[utils.key(t,'a')]    -- Find 'a'                y == 42
    z = t[utils.key(t,'A')]    -- Find 'A' (same as 'a')  z == 42
    m = t[utils.key(t,'b')]    -- Show 'b' is missing     m == nil
    k = utils.key(t,'s')       -- Which 's/S' is first?   k is indeterminate
    l = t['s']                 -- Get 's' and 'S' in      l = 37
    u = t['S']                 --   the usual manner.     u = 73
    t[utils.key(t,'a')] = nil  -- Delete 'a'              t == {s=37, S=73}
    ```

    As implied by the `k = ` example above, lua associative tables are unordered, so there's no guarantee that 's' (or 'S') is the first one to be found. In this context, that shouldn't be too much of an issue.

2. Use this new function to find the 'Host' and 'Content-Type' headers' values no matter in what case they were defined.


ShivamJoker:
@teto we please get this merged? I struggled a lot with this case sensitivity.

NTBBloodbath (approved):
LGTM! Sorry for the delay, life is a little busy and I can't fully focus on every project I have yet!

PR #239, opened 2023-10-02
Fix finding the start and end of the current query.

### LINKED ISSUE #173: wrong query selected in multiple query file
With multiple queyr files, sometimes the wrong query is used (i.e., a one different from the query the cursor points at), for instance I can have the cursor within the query 1 paragraph
```
query1
---
query2
```
but query2 will be sent

A similar issue  https://github.com/rest-nvim/rest.nvim/issues/172 . 

I want to invest some time to see if I can fix it with the treesitter grammar but not sure when (hope someone beats me to it).

PhilRunninger:
Fixes #173.

This isn't a treesitter solution, but it's at least a quick and easy, if only temporary, patch.

1. start_request was searching forward from the cursor position. Adding 'b' to the search flags fixed that.
2. Both start_request and end_request depended on the user having set 'nowrapscan' to prevent wrapping around the end of the file. Such wrapping could result in finding the wrong query. Adding 'W' to the search flags fixed that, and removed the need for the stopline parameter.
3. Refactored the way last_line is used in end_request.

### DIFF
--- a/lua/rest-nvim/request/init.lua
+++ b/lua/rest-nvim/request/init.lua
@@ -193,7 +193,7 @@ local function start_request(bufnr, linenumber)
   local oldlinenumber = linenumber
   utils.move_cursor(bufnr, linenumber)
 
-  local res = vim.fn.search("^GET\\|^POST\\|^PUT\\|^PATCH\\|^DELETE", "cn")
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):
Thank you so much for this patch. It is indeed a great solution in the meantime :)

PR #243, opened 2023-10-07
feat(doc): use treesitter injection

zirbest:


NTBBloodbath (approved):
Thanks, this will definitely improve the documentation look!

PR #257, opened 2023-11-19
feat: stay in current window on split

volomas:
When you run request for this first time, focus is getting switched to the new windows with results buffer.
I've added simple flag that allows to configure this behavior: either stay in current windows (.http file) or change to results window (default)

### DIFF
--- a/lua/rest-nvim/config/init.lua
+++ b/lua/rest-nvim/config/init.lua
@@ -3,6 +3,7 @@ local M = {}
 local config = {
   result_split_horizontal = false,
   result_split_in_place = false,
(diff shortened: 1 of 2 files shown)


NTBBloodbath (approved):
Looks good to me, thank you very much for this new addition!

Side note: maybe we should make this the default behavior afterwards?

volomas:
> Looks good to me, thank you very much for this new addition!
> 
> Side note: maybe we should make this the default behavior afterwards?

As for me, I would prefer to have it default.
Decided not to do so in this PR, because that's not what existing users might be used to.

Issue #256, opened 2023-11-16
Add statistics for requests


Rasmus-Bertell:
How long the request took to execute and the response size in bytes are something I usually look at when testing APIs and would be a welcome addition to this plugin in my opinion.

I might be willing to look into implementing this, but don't know if I have the time to do it.

Rasmus-Bertell:
Started exploring this but cUrl really doesn't make it easy to have timings while preserving the output. The method I came up with is to add the timings as a separate row after output with `-w "\n%{time_total}..."` and then remove it before handling the output.

PR #260, opened 2023-11-21
feat: refactor writing to buffer

Rasmus-Bertell:
Makes writing to the buffer cleaner and easier to add more blocks.
This is pre work for #256 so that it's easier to add the statistics block into the output.

This shouldn't affect the output other than it removes the line `#+END` between the command and headers on subsequent requests if the buffer wasn't closed. It was leftover since the buffer wasn't cleared properly.


NTBBloodbath (approved):
This is awesome, thank you so much. The codebase is much more readable now for my eyes and I hope it is for others too. Also fixed a small bug along the way so excellent :)

PR #267, opened 2023-12-05
feat: add buffer_name option

davidsierradz:
Hi! It would be nice to have the ability to keep multiple result buffers open at the same time, like in [diepm/vim-rest-console](https://github.com/diepm/vim-rest-console#42-multiple-vrc-buffers), I'm guessing adding an option would be the most flexible way.

You could check a buffer variable like in vim-rest-console:

```lua
buffer_name = function()
  return vim.b.rest_nvim_result_buffer or 'rest_nvim_results'
end
```

Thanks, let me know any questions or review' comments.

### DIFF
--- a/lua/rest-nvim/config/init.lua
+++ b/lua/rest-nvim/config/init.lua
@@ -33,6 +33,9 @@ local config = {
         }, body):gsub("\n$", "")
       end,
     },
+    buffer_name = function()
+      return 'rest_nvim_results'
+    end,
   },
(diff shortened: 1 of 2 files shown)


boltlessengineer:
I’m a bit afraid to allow `buffer_name` option. As now rest.nvim creates multiple buffers for each result UI panes, allowing custom buffer name might allow users to break things unintentionally.
We can support multiple result buffers in first place with response saving feature (#359)

PR #276, opened 2024-01-26
fix: return non json body

purefun:
fix for non json body like form data:

```http
POST https://httpbin.org/post
Content-Type: application/x-www-form-urlencoded

a=1&b=2
```


NTBBloodbath (approved):
Woops, nice catch!

Thank you for the patch :)

PR #278, opened 2024-01-31
add telescope extension to select env file

BingCoke:


NTBBloodbath:
Hey, the truth is I never expected someone to do something like this and I love the idea, thank you very much!

I'm currently rewriting the entire plugin and I'd like this to be in the `dev` branch to include in the next version.

Also I saw this PR and decided to expose a function `("rest-nvim.functions").find_env_files()` in the rewrite branch which I think could be very useful and would reduce the telescope extension code, also removing the optional dependency on `fd` :)

BingCoke:
I think it could be better used as a extension as other plugin like project.nvim so that optional to load it.
I also want to add other feature like 
1. add lualine source, when we in http file we can check what env we select and add EditEnv command to edit current env file
2. add http file snippets
3. add attach function (you know there is not a http lsp) we can keybinding just in http file

Do you think these features could be added to this repository?

NTBBloodbath:
Woops, I just woke up and have some time to look at GitHub.

Of course, why not? Those ideas sound super useful, I will be happy to have them in the repository :p

Otard95:
This can be fairly easily accomplished in your own dotfiles as well. Here's [my config](https://github.com/Otard95/dotfiles/blob/1898c1d6177087827e8db23aeae2a65445e2d02d/config/nvim/lua/plugins/rest.lua#L2-L25) for rest.nvim, hope it helps someone.

BingCoke:
> This can be fairly easily accomplished in your own dotfiles as well. Here's [my config](https://github.com/Otard95/dotfiles/blob/1898c1d6177087827e8db23aeae2a65445e2d02d/config/nvim/lua/plugins/rest.lua#L2-L25) for rest.nvim, hope it helps someone.

yes of course,but i think it could be better be managed by this repo other than user's config.

Otard95:
> yes of course,but i think it could be better be managed by this repo other than user's config.

I didn't mean to imply that it shouldn't, sorry if it came across that way. Only wanted to give an option to anyone else coming along in the mean time 😊

NTBBloodbath:
Sorry it took so long for me to merge this. I've been somewhat busy with other projects and the `rest.nvim` incoming version is also a priority, I'll make sure to cherry-pick this and the lualine features into the `dev` branch, thank you so much :)

PR #280, opened 2024-02-02
fix bug when headers have  `host`

mtfcd:
in the original code, the `host` header will add to the url. maybe it allows the url part only have the path part. but if it’s a complete url, it will form a invalid url.
so the change is : add host before url only if it starts with '/', otherwise it probably starts with 'http'.


NTBBloodbath (approved):
Thank you so much :)

PR #285, opened 2024-02-18
Telescope extension can select env files ignored by git

### LINKED ISSUE #283: Unable to find env files in the env selector if they are defined in gitignore
Noticed after adding *.env to .gitignore, that the selector stopped finding my environments.
Removing it again, and the environments show up in the picker.

Could you add -I as fd argument so telescope select env extension works for files in .gitignore?
Alternatively, some other way, or an example of how to make it work.

Just in case it helps,
my env files are named:
prod.env
beta.env

lhanson:
When trying to select an environment file with the Telescope extension, the underlying `fd -H` command would match hidden files, but not files that are matched by `.gitignore`. Adding the `-I` option matches files that are ignored by git.

Fixes #283

### DIFF
--- a/lua/telescope/_extensions/rest.lua
+++ b/lua/telescope/_extensions/rest.lua
@@ -20,7 +20,7 @@ local function rest_env_select(opt)
   local pattern = config.get("env_pattern")
   local edit = config.get("env_edit_command")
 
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):
Thank you so much 💜

Issue #184, opened 2023-02-26
filetype not applying


dimas-cyriaco:
I have just installed the plugin and the filetype is not been applied automatically.

If I set the filetype manually everything works perfectly fine.

I'm using [LazyVim](https://github.com/LazyVim/LazyVim) and the plugin configured like this:

```lua
return {
  "rest-nvim/rest.nvim",
  dependencies = { "nvim-lua/plenary.nvim" },
  keys = {
    { "<leader>rr", "<Plug>RestNvim", desc = "Run request" },
  },
  config = function()
    require('rest-nvim').setup({
      -- Open request results in a horizontal split
      result_split_horizontal = false,
      -- Keep the http file buffer above|left when split horizontal|vertical
      result_split_in_place = false,
      -- Skip SSL verification, useful for unknown certificates
      skip_ssl_verification = false,
      -- Encode URL before making request
      encode_url = true,
      -- Highlight request on run
      highlight = {
        enabled = true,
        timeout = 150,
      },
      result = {
        -- toggle showing URL, HTTP info, headers at top the of result window
        show_url = true,
        show_http_info = true,
        show_headers = true,
        -- executables or functions for formatting response body [optional]
        -- set them to false if you want to disable them
        formatters = {
          json = "jq",
          html = function(body)
            return vim.fn.system({ "tidy", "-i", "-q", "-" }, body)
          end
        },
      },
      -- Jump to request line on run
      jump_to_request = false,
      env_file = '.env',
      custom_dynamic_variables = {},
      yank_dry_run = true,
    })
  end
}
```

Tried with an empty setup too, but no luck.

piotr-serafin-red:
Hi, just add ft to the table.

```lua
return {
  "rest-nvim/rest.nvim",
  ft = "http",
  dependencies = { "nvim-lua/plenary.nvim" },
  keys = {
    { "<leader>rr", "<Plug>RestNvim", desc = "Run request" },
  },
  config = function()
  (...)
```
Ofcourse file with your request should end with .http.

PR #288, opened 2024-02-23
docs: add `ft` to `lazy.nvim` config example

lhanson:
Without this, the filetype is not applied correctly.
#184


NTBBloodbath (approved):
Thanks for the patch :)

Issue #294, opened 2024-03-09
bug(v2): error when cannot get the result's content-type


skyf0cker:
when cannot get the content type from the response, current implement will throw error
<img width="677" alt="image" src="https://github.com/rest-nvim/rest.nvim/assets/32331521/5e1b715f-289c-412e-bac6-249af316ee26">

<img width="457" alt="image" src="https://github.com/rest-nvim/rest.nvim/assets/32331521/2f5f164e-4418-4546-93fd-574278b9c17f">

got error:
```plain
Error executing Lua callback: /Users/dd/.luarocks/share/lua/5.1/nio/tasks.lua:95: Async task failed without callback: The coroutine failed with this message: 
.../share/nvim/lazy/rest.nvim/lua/rest-nvim/result/init.lua:237: attempt to concatenate local 'res_type' (a nil value)
stack traceback:
	.../share/nvim/lazy/rest.nvim/lua/rest-nvim/result/init.lua: in function 'format_body'
	.../share/nvim/lazy/rest.nvim/lua/rest-nvim/result/init.lua:304: in function <.../share/nvim/lazy/rest.nvim/lua/rest-nvim/result/init.lua:303>
stack traceback:
	[C]: in function 'error'
	/Users/dd/.luarocks/share/lua/5.1/nio/tasks.lua:95: in function 'close_task'
	/Users/dd/.luarocks/share/lua/5.1/nio/tasks.lua:117: in function 'step'
	/Users/dd/.luarocks/share/lua/5.1/nio/tasks.lua:145: in function 'run'
	.../share/nvim/lazy/rest.nvim/lua/rest-nvim/result/init.lua:303: in function 'write_res'
	...al/share/nvim/lazy/rest.nvim/lua/rest-nvim/functions.lua:100: in function 'exec'
	...cal/share/nvim/lazy/rest.nvim/lua/rest-nvim/commands.lua:44: in function 'impl'
	...cal/share/nvim/lazy/rest.nvim/lua/rest-nvim/commands.lua:182: in function <...cal/share/nvim/lazy/rest.nvim/lua/rest-nvim/commands.lua:165>
	```

PR #295, opened 2024-03-09
fix: ignore error log and adding syntax highlights for result when can not get the res_type

skyf0cker:
fix https://github.com/rest-nvim/rest.nvim/issues/294


NTBBloodbath (approved):
Woops, the truth is that this bug would never have crossed my mind. Thank you very much for sending a fix :D

skyf0cker:
> Woops, the truth is that this bug would never have crossed my mind. Thank you very much for sending a fix :D

Thanks for your reply :) But it looks like the lint check in CI broke.

NTBBloodbath:
Oh, yeah. I think we can ignore it, I didn't find anything very relevant. I have to adjust the luacheck configuration to the new codebase changes :)

PR #296, opened 2024-03-11
Telescope extension Doc dev

ZieMcd:
Added a bit to documentation about getting telescope extension to work on Ubuntu and Debian systems


NTBBloodbath (approved):
Very appreciated :)

PR #297, opened 2024-03-13
Get json content type for custom content types

ggsalas:
I have found a bug using the contentful api. This api returns this content type: 
```
Content-Type: application/vnd.contentful.delivery.v1+json
```
Currently the content type is `vnd` instead of `json`. 

I have added this fix, but I think can be improved  extending to all the content types this plugin want to support, probably are only 3: json, html and txt. let me know what you think, and thanks for this great plugin 🎉

### DIFF
--- a/lua/rest-nvim/curl/init.lua
+++ b/lua/rest-nvim/curl/init.lua
@@ -133,7 +133,13 @@ local function create_callback(curl_cmd, opts)
 
     local content_type = res.headers[utils.key(res.headers, "content-type")]
     if content_type then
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):
Hey, thank you so much for this patch!

> I think can be improved extending to all the content types this plugin want to support, probably are only 3: json, html and txt.

Of course, in fact, I think this little bottleneck was already _or can be easily_ fixed in the `dev`elopment branch (which is `rest.nvim`'s v2, that is close to release) thanks to depending on [mimetypes.lua](https://luarocks.org/modules/lunarmodules/mimetypes) :)

PR #302, opened 2024-03-18
fix: ensure help window width is an int

PriceHiller:
Copied from the body of the commit:

> Problem:
> 
> The help window causes an error when calling `vim.api.nvim_open_win` because the width being passed is not an integer.
> 
> Neovim's `vim.api.nvim_open_win` function requires that the width and height be specified in whole integers.
> 
> The width is being calculated through a division operation causing it to potentially be a number that is not an int.
> 
> Solution:
> 
> Floor the calculation of the help window width so the width is always rounded down to the nearest integer and can thus be used with `vim.api.nvim_open_win`.

### DIFF
--- a/lua/rest-nvim/result/help.lua
+++ b/lua/rest-nvim/result/help.lua
@@ -62,7 +62,7 @@ function help.open()
   end
 
   -- Help window sizing and positioning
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):
Thank you so much for this patch :D

PR #304, opened 2024-03-18
fix(parser): make `parse_request` handle `request` TS nodes

PriceHiller:
Copied from the body of the commit:

> Problem:
> 
> Placing cursor between a request method and a uri and then invoking `:Rest run` causes the node to be incorrectly parsed leading to an error.
> 
> See the following where the column above the caret (^) denotes the cursor position:
> 
> ```
> GET https://google.com
>    ^ cursor above caret
> ```
> 
> Running this will cause an error. This is because the `parser.parse_request` method cannot handle Treesitter nodes of type `request`.
> 
> Solution:
> 
> Make the `parser.parse_request` function handle Treesitter nodes of type `request` by recursively calling itself on the `request` node and taking that output when it finds a `request` node.

Example of the problem with a recording:
rest-nvim-parser-bug

This problem implies a deeper issue to my eyes as I noticed there's a function `parser.look_behind_until` that's supposed to get the top level `request` node I assume. I don't know enough about the code of `rest.nvim` to make a strong statement beyond this unfortunately 🙁.

I'm making this PR because it certainly solves the problem noted above, but I don't know if it's the *best* solution. If you know a better way of doing this please advise, I'll gladly close this PR and fix the underlying issue with some guidance (assuming there actually is an underlying issue).

### DIFF
--- a/lua/rest-nvim/parser/init.lua
+++ b/lua/rest-nvim/parser/init.lua
@@ -221,6 +221,8 @@ function parser.parse_request(children_nodes, variables)
     elseif node_type == "http_version" then
       local http_version = assert(get_node_text(node, 0))
       request.http_version = http_version:gsub("HTTP/", "")
+    elseif node_type == "request" then
+      request = parser.parse_request(traverse_request(node), variables)
     end
   end
 


NTBBloodbath (approved):
Woops, nice catch and thanks again for submitting a patch! 😅

I think this way is fine, I feel that using the `look_behind_until` function could bring more unnecessary complexity , which is not something we want.

Ps: in case you feel the code could be documented further or things like that please let me know, that's a very important point :)

PriceHiller:
> Ps: in case you feel the code could be documented further or things like that please let me know, that's a very important point :)

As a heads up when I mentioned "I don't know enough about the code of `rest.nvim`" I meant that as in I literally looked at the code for no more than 10 minutes total to fix the some issues I encountered. I don't want to come off as "your code is hard to read and undocumented" -- not the intent at all.

I imagine more stuff will crop up from the `2.0.0` release and I'll dive into it more. Once I get more comfortable with the code then maybe I'll look back at improved documentation and whatnot if that actually is an issue at all.

For now, just at a glance, it looks pretty solid to my eyes.

The plugin kicks ass keep it up ❤️.

PR #325, opened 2024-03-21
fix(run last): not modify body in order to use it again

### LINKED ISSUE #324: bug: modifying body makes `Rest run last` not work as expected
Currently, the body is modified in order to remove the `__TYPE` field before encoding it. This makes `Rest run last` fail when a `__TYPE` field is expected.

TheLeoP:
Fixes #324


NTBBloodbath (approved):
Hi, thanks for spotting and fixing this bug! :)

PR #330, opened 2024-03-22
docs(README): update instructions to adapt to latest `luarocks.nvim` changes

vhyrro:
This PR updates the rest.nvim README with the newest changes related to luarocks.nvim. Once this gets merged I'll also merge the luarocks.nvim `go-away-python` branch into `main`.

The `go-away-python` branch removes the need for python as an external dependency, completely eliminates hererocks, fixes several bugs on different platforms and generally tries to provide a more streamlined experience. That does come unfortunately at the cost of a breaking change, but such is life sometimes :)

 This will probably be the last breaking change luarocks.nvim has in a *while*.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,26 +67,13 @@ CLI. For more information on this, please see this [blog post](https://amartin.c
 :Rocks install rest.nvim
 ```
 
-### [packer.nvim](https://github.com/wbthomason/packer.nvim)
-
-```lua
-use {
-  "rest-nvim/rest.nvim",
-  rocks = { "lua-curl", "nvim-nio", "mimetypes", "xml2lua" },
-  config = function()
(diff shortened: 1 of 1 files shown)


vhyrro:
Extra notice: I've recognized that a lot of users erroneously put their rock list in the `luarocks.nvim` setup function, even though they don't *have* to. Rest.nvim's build.lua handles everything automatically.

For this reason I also put the lazy.nvim instructions above the packer.nvim instructions. Should prevent most of these errors :laughing:

vhyrro:
Note: The changes have landed, so merging this sooner or later would be awesome :)

NTBBloodbath (approved):
Thanks very much mate :)

Issue #339, opened 2024-03-26
Error on winbar highlighting result


kevinm6:
Hello, thanks for the plugin.
Using it and updated to the last release with `luarocks` dependency and the other updates I get an error on the result highlight of the winbar.

Due to this part of the code
https://github.com/rest-nvim/rest.nvim/blob/5300ae0b111dbee1322949a197cc424eac6b849f/lua/rest-nvim/result/winbar.lua#L64-L69

when the Hex value of the highlight contains two prefixed zeros (example `#00FF87`) it doesn’t match to any of the if statement since you’re checking if the chars left after the format are **5**, but in this case they are **4**.

In the most simple way is enough to do another check and have something like this

```lua
  local hl_fg = string.format("%02X", vim.api.nvim_get_hl(0, { name = name, link = false }).fg)
  if #hl_fg == 4 then
    hl_fg = "00" .. hl_fg
  elseif #hl_fg == 5 then
    hl_fg = "0" .. hl_fg
  end
  hl_fg = "#" .. hl_fg
  return hl_fg
```

PR #340, opened 2024-03-26
fix(functions): handle highlights of winbar result with prefixed zeros

kevinm6:
This should avoid errors as reported on issue #339


NTBBloodbath (approved):
Hey, it seems that this is an edge case I didn't catch before. Thank you very much for reporting it and submitting a patch :)

PR #342, opened 2024-03-27
ref(winbar): pad the hex with 6 zeroes using string.format

tamton-aquib:
Cleaner way for padding with zeroes.
Related to: #339


NTBBloodbath (approved):
Thank you mate! :)

PR #351, opened 2024-04-06
feat: custom mappings for cycle result pane

daniilrozanov:
There was added new table in `result` table named `keybinds`. With this table you can redefine mapping for `cycle_result_pane` action. To do this, in config you need to add
```lua
result = {
  keybinds = {
    prev = "J", -- or any
    next = "K", -- or any
  }
}
```

Other feature is ability to choose does these mappings attach to request result buffer or not. If you want to make mappings locally in buffer, add
```lua
result = {
  keybinds = {
    -- ...
    buffer_local = true
  }
}
```
This feature is useful if you already have other actions mapped with `H` or `L`.

Defaults for `result.keybinds` are following:
```lua
keybinds = {
  buffer_local = false,
  prev = "H",
  next = "L",
}
```
So plugin acts like before if `keybinds` is not set.

### DIFF
--- a/lua/rest-nvim/config/init.lua
+++ b/lua/rest-nvim/config/init.lua
@@ -126,6 +126,11 @@ local default_config = {
         end,
       },
     },
+    keybinds = {
+      buffer_local = false,
+      prev = "H",
+      next = "L",
+    },
   },
   highlight = {
     enable = true,
--- a/lua/rest-nvim/config/init.lua
+++ b/lua/rest-nvim/config/init.lua
@@ -20,6 +20,7 @@ local logger = require("rest-nvim.logger")
 ---@class RestConfigResult
 ---@field split RestConfigResultSplit Result split window behavior
 ---@field behavior RestConfigResultBehavior Result buffer behavior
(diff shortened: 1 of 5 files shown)


lua/rest-nvim/config/init.lua
```diff
@@ -126,6 +126,11 @@ local default_config = {
         end,
       },
     },
+    keybinds = {
+      buffer_local = false,
+      prev = "H",
+      next = "L",
+    },
```
NTBBloodbath, line 133: Could you please annotate this configuration options table? The rest of the annotations are at the top of the file, in case you need some examples and where to place them

NTBBloodbath (changes requested):

lua/rest-nvim/config/init.lua
daniilrozanov: [Done](https://github.com/rest-nvim/rest.nvim/pull/351/commits/e2e77cd50fcb0946126b9c35b9434d937ac2706a)

NTBBloodbath (approved):
Everything looks fine now, thank you so much! :)

countcain:
thank you so much!

PR #354, opened 2024-04-10
fix: correctly urlencode query parameters

### LINKED ISSUE #317: Encoding of = in querystring
Hi,

I have a strange bug with query strings.

This is my http file:

```
GET https://[redacted/path/to/something/?category=tests
Accept-Language: fr
```

And if I run it, the `=` gets encoded as `%3d` and the request fails:

```
#+RES
array(1) {
  ["category=tests"]=>
  string(0) ""
}

#+END
```

Running with curl sends the request correctly and the array is populated as expected.

flexagoon:
Fixes #317 

Splits the query into individual parameters and appends each one to the url using the `U_APPENDQUERY` flag. Handling of the `=` character is not needed due to how the flag works:

> When [CURLU_APPENDQUERY](https://curl.se/libcurl/c/curl_url_set.html#CURLUAPPENDQUERY) is used together with [CURLU_URLENCODE](https://curl.se/libcurl/c/curl_url_set.html#CURLUURLENCODE), the first '=' symbol is not URL encoded.

See https://curl.se/libcurl/c/curl_url_set.html#CURLUAPPENDQUERY

### DIFF
--- a/lua/rest-nvim/client/curl.lua
+++ b/lua/rest-nvim/client/curl.lua
@@ -204,7 +204,13 @@ function client.request(request)
       local _url = curl.url()
       _url:set_url(request.request.url)
       -- Re-add the request query with the encoded parameters
-      _url:set_query(_url:get_query(), curl.U_URLENCODE)
+      local query = _url:get_query()
+      if type(query) == "string" then
+        _url:set_query('')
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):
Looks good to me, thank you very much for the patch! :)

PR #358, opened 2024-04-13
fix: change default HL behavior

Redoxahmii:
Instead of global keymaps only HL for switching tabs when focused on that buffer as the default behavior.


NTBBloodbath (approved):
Thank you so much :)

Redoxahmii:
My pleasure :+1:

PR #365, opened 2024-04-17
fix: match lowercase content-type header

disrupted:
HTTP headers are case insensitive. However, rest.nvim failed parsing lowercase `content-type` header


NTBBloodbath (commented):
Thank you so much for addressing this! :D

PR #377, opened 2024-05-03
refactor: remove luanil opts for vim.json.decode

aktersnurra:
This PR removes the luanil opts from the `vim.json.decode`, since when they are provided all instances of key-value pairs with null values are removed, see my issue in [neovim](https://github.com/neovim/neovim/issues/28619). lua-tables cannot hold nil values:

```
Like indices, the value of a table field can be of any type (except `nil`). In
particular, because functions are first-class values, table fields may contain
functions. Thus tables may also carry methods (see |lua-function-define|).
```

which is the root cause to this behavior.


NTBBloodbath (approved):
Thanks for addressing this and sorry for the inconveniences! :)

PR #379, opened 2024-05-04
docs: add alternative for selecting environment

yujinyuz:
This lets non-telescope users select the environment with better ux


NTBBloodbath (approved):
Thank you! :D

PR #392, opened 2024-06-03
chore: add `doc/tags` to `.gitignore`

eduardomezencio:
Just adding doc/tags to .gitignore so that when tags are generated with `:helptags` they don't make the index dirty, specially for people who use cloned git repos or git submodules for plugin management.

This is also good to make sure no outdated tags are shipped with the project. The current commited tags file was outdated, for example.


teto (approved):

NTBBloodbath (approved):

NTBBloodbath:
Thank you so much :)

PR #401, opened 2024-08-12
v3 refactor

boltlessengineer:
tl;dr: Rewrite everything except dotenv parser.

### Noticeable changes:

- uses pure `curl` instead of `libcurl` to get non-blocking asynchronous request
- most user command won’t be compatible with v2
- most apis won’t be compatible with v2
- most config options won’t be compatible with v2
- pre-request & request-handler scripts won’t be compatible with v2
- uses [tree-sitter-http] parser for response view (so now you can fold/unfold the response body as you want)
- now plugin works without `setup()` call
- `:Rest` command is registered globally. You can run `:Rest last` on any buffer

### New features

- cookies support

### More todos

- [x] v2 compatibility (mostly deprecation notice)
- [x] rewrite documents
- [x] cleanup configs
- [x] rewrite telescope picker
- [x] add more tests

[tree-sitter-http]: https://github.com/rest-nvim/tree-sitter-http


gorillamoe:
Congrats :rocket: !

NTBBloodbath:
question: is there going to be support for post-response actions? I was thinking about making an external integration for SQL databases to run info queries after the request and such things so I'm wondering if that will be still possible :p

boltlessengineer:
> is there going to be support for post-response actions?

@NTBBloodbath Yes it still supports that thing (post-response action / post-request action / response-handler script / or whatever you call it)

User can do either:

- put custom script in http file like this [example](https://github.com/rest-nvim/rest.nvim/blob/6a7699bae02caf5bc98e229fd1216aa16a3e7bb3/spec/examples/script/post_request_script.http) (the `{% … %}` syntax, local to each request)
- make a lua autocommand in their editor to alter data before and after the request.

I’m also planning to allow external script runner clients but it will be post-v3-release goal.

README.md
```diff
@@ -74,39 +60,17 @@ CLI. For more information on this, please see this [blog post](https://amartin.c
...
 ### [packer.nvim](https://github.com/wbthomason/packer.nvim)
 
 ```lua
 use {
   "rest-nvim/rest.nvim",
-  rocks = { "lua-curl", "nvim-nio", "mimetypes", "xml2lua" },
-  config = function()
-    require("rest-nvim").setup()
-  end,
+  rocks = { "nvim-nio", "mimetypes", "xml2lua", "fidget.nvim" },
```
Alienover, line 73: Wondering why we don't need these dependencies when installing by `lazy.nvim`?

boltlessengineer: afaik lazy.nvim now supports luarocks dependencies, so as `rest-nvim` is on luarocks, lazy.nvim can know all needed dependencies.
For example, [`neorg` also don't need to write dependencies when installing by `lazy.nvim`](https://github.com/nvim-neorg/neorg?tab=readme-ov-file#lazynvim)

PR #407, opened 2024-08-22
fix(parse http time): update lua pattern to support hyphen separated date

aktersnurra:
This patch updates the regex pattern for the time_str to support the format `Tue, 18-Feb-2025 09:59:46 GMT`

### DIFF
--- a/lua/rest-nvim/utils.lua
+++ b/lua/rest-nvim/utils.lua
@@ -84,7 +84,7 @@ function utils.read_file(path)
 end
 
 function utils.parse_http_time(time_str)
-    local pattern = "(%a+), (%d+)[\\s-](%a+)[\\s-](%d+) (%d+):(%d+):(%d+) GMT"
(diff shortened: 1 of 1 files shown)


lua/rest-nvim/utils.lua
```diff
@@ -84,7 +84,7 @@ function utils.read_file(path)
 end
 
 function utils.parse_http_time(time_str)
-    local pattern = "(%a+), (%d+) (%a+) (%d+) (%d+):(%d+):(%d+) GMT"
+    local pattern = "(%a+), (%d+)[\\s-](%a+)[\\s-](%d+) (%d+):(%d+):(%d+) GMT"
```
boltlessengineer, line 87: `pattern` here is lua pattern, not regex, therefore `\s` won’t work. You can use `%s` instead.

boltlessengineer (changes requested):

Issue #435, opened 2024-09-02
Feature: set curl opts globally/per request


sheimer:
### Issues

- [X] I have checked [existing issues](https://github.com/rest-nvim/rest.nvim/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

I have a route which sends a csv file with content-encoding gzip.

To be able to see the result in neovim, I have to set the `--compressed` arg on the curl command. I found no way to do so but changing the args in `lua/rest-nvim/client/curl/cli.lua`.

Maybe there is a way to allow args to be passed to curl in general? Or set --compressed, if Accept-Encoding header is set and contains gzip?

boltlessengineer:
I’m thinking of providing `meta` field for `rest.Request` type.
So users can make `User RestRequestPre` autocommand to update the request metadata before running a request
and clients like `curl` can use that metadata on request.

Introducing additional document comment (e.g. `# @curl-args --compressed`) might make things easier, but I’m bit skeptical to introduce too much new specs on top of other plugins from intellij/vscode)

Adding `--compressed` argument automatically based on header type sounds like a good idea.

Would this fit your needs?
Also please tell me if you are willing to make a PR. If not, I will do it but it might take some time due to personal schedules.

sheimer:
I will try if I can do a PR...time might be a problem, and maybe my capabilities as well ^^ We will see :)

boltlessengineer:
#439 is merged now. manually closing the issue

PR #439, opened 2024-09-03
feat: use curl --compressed argument if Accept-Encoding includes gzip

sheimer:
Adds `--compressed` argument automatically, if header `Accept-Encoding` is present and includes `gzip`.

A new builder method builder.extras is used to add the argument.

Solves problem https://github.com/rest-nvim/rest.nvim/issues/435

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ local default_config = {
     },
     ---@class rest.Config.Response
     response = {
+        ---Default response hooks
         ---@class rest.Config.Response.Hooks
         hooks = {
             ---@type boolean Decode the request URL segments on response UI to improve readability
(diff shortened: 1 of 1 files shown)


boltlessengineer:
@sheimer I added `clients.curl.opts.set_compressed` option which is `false` by default.
You can enable this option to set `--compressed` argument when `Accept-Encoding` header contains `”gzip”`.

If you are fine with this opt-in implementation, I’ll merge the PR.

sheimer:
I am totally fine with this opt-in option on the client! Thank you very much :)

boltlessengineer:
Oops I messed up the branches while rebasing it. This is merged to main now.

PR #446, opened 2024-09-05
fix: scope and buf number in setting options

Alienover:
image


boltlessengineer:
Thank you for your contribution! I'm blindly merging it (outside now so I can't test). I wonder why tests couldn't catch this issue. Maybe it is caused by different neovim version?

Alienover:
https://github.com/neovim/neovim/commit/743c5808b616daece73145ddd12b39ae07bbd87a
Saw the options validation was updated 3 months ago, which is still in the `nightly` version.

Issue #459, opened 2024-09-14
Env file is not loaded if the file does not end in .env


sgatu:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

v0.10.1

### Operating system/version

Ubuntu 23.10

### Actual behavior

After selecting the env file I get this.

`Error executing vim.schedule lua callback: ...local/share/nvim/lazy/rest.nvim/lua/rest-nvim/dotenv.lua:36: path: expected `.env` or `.json` filetype, got .env.local`

### Expected behavior

The env file to be loaded and env variables to be available.

### Steps to reproduce

Execute :Rest env select
Select a file that does not end in .env, like the error above ending in .local

### Other information

From the documentation it looks like it should be able to accept files with different extension
``` 
env = {
        ---@type boolean
        enable = true,
        ---@type string
        pattern = ".*%.env.*",
}    
```
And from what I'm seeing in the code, in [lua/rest-nvim/parser/dotenv.lua](https://github.com/rest-nvim/rest.nvim/blob/main/lua/rest-nvim/parser/dotenv.lua), the library tries parse both json and env file, separating the parsing process by extension. Yet the documentation does not talk of the possibility of a json env file.

PR #460, opened 2024-09-14
fix: remove env file extension validation

sgatu:
Now if it ends in json it will try to parse it as json, else it will fallback to .env file parsing. 

Fixes issue #459


boltlessengineer:

PR #475, opened 2024-09-28
fix: handle when format target is empty string (# 473)

boltlessengineer:
reference: #473


jrebs:
Thanks for trying to solve the problem I brought up in that PR, but this change doesn't appear to fix it. I have updated to 3.8.2 and when I hit that endpoint with a knowingly bad Bearer token, I get the big red ocean of error stack information. When I comment out utils.lua lines 329-331, I am presented with a usable 401 response. The gggqG call still seems to be blowing this up.

boltlessengineer:
@jrebs Can you open a separate issue so I can keep track of this there? That will be helpful.

I want to make sure exactly what is causing `gq` to fail.

PR #499, opened 2024-12-19
docs: added lazyvim reference configs

jsouzadasilva:
This adds a section called "Reference configs", under "Usage". I added the reference for LazyVim, but this could be expanded for other neovim distributions.

The reference config can give people something to start with, or even use it as is.

Let me know if I should do any formatting of it's a bad idea to add this in the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,115 @@ response.body = vim.json.encode(json)
 Put `# @lang=lua` comment just above any script elements.
 Scripts without `@lang` will be parsed as javascript code to match with [http spec](https://www.jetbrains.com/help/idea/exploring-http-syntax.html#response-handling).
 
+### Reference configs
+
+The configs in this section are to be used as a reference, you may use it as is, or modify it to suit
+your needs.
+
+#### LazyVim
+
+<details>
+  <summary>plugins/rest-nvim.lua</summary>
+
+```lua
+return {
+  {
+    "rest-nvim/tree-sitter-http",
+  },
+  {
+    "nvim-treesitter/nvim-treesitter",
+    opts = {
+      ensure_installed = {
+        "http",
+      },
+    },
+  },
+  {
+    "folke/which-key.nvim",
+    event = "VeryLazy",
+    opts_extend = { "spec" },
+    opts = {
+      spec = {
+        {
+          mode = { "n", "v" },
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -252,6 +252,112 @@ response.body = vim.json.encode(json)
...
+              vim.notify("No input provided", vim.log.levels.ERROR) -- Show error
+            end
+          end)
+        end,
+        desc = "Register .env file to current .http file",
+      },
+    },
+  },
+}
+```
```
NTBBloodbath, line 359: nitpick: enclosing the reference configs section children headers in an HTML `<details>` tag could reduce the visual clutter by default, If anyone wants to see them, they can use the toggle "button" text from the tag.

jsouzadasilva: addressed, thank you!

boltlessengineer:
Thanks for the PR and for taking the time to contribute! While I appreciate the effort, I think adding reference configs to the README isn’t the best fit for rest.nvim.

- rest.nvim already use rockspec to manage dependencies so you don't need to define those in lazyspec (especially tree-sitter-http. If you are using current version of nvim-treesitter, adding http parser in nvim-treesitter config is enough)
- `<leader>` keymaps are usually for user-defined ones, so including defaults could unintentionally override user preferences.

I’d like to keep the README focused on core usage and let users adapt the plugin to their setups.

Maybe we can add this in wiki instead.

boltlessengineer (commented):
I don't think it is good idea to add this in Readme. Maybe we can add this in wiki instead.

jsouzadasilva:
> Thanks for the PR and for taking the time to contribute! While I appreciate the effort, I think adding reference configs to the README isn’t the best fit for rest.nvim.
> 
> * rest.nvim already use rockspec to manage dependencies so you don't need to define those in lazyspec (especially tree-sitter-http. If you are using current version of nvim-treesitter, adding http parser in nvim-treesitter config is enough)
> * `<leader>` keymaps are usually for user-defined ones, so including defaults could unintentionally override user preferences.
> 
> I’d like to keep the README focused on core usage and let users adapt the plugin to their setups.
> 
> Maybe we can add this in wiki instead.

Got it, I'm closing this then. If you need help adding this to the wiki let me know, thanks!

PR #516, opened 2025-01-22
Add a javascript runner for running javascript before and after requests

mrloop:
This implement a basic javascript runner.
* support Response API
* support set/get Request API variables
* support set/get client global variables
* support jsonPath query in javascript

Dev notes:
A intermediate file is created for each request in the script folder, `last_javascript.mjs` is the last javascript file that was run. This is useful for debugging. For instance you can run the file from the command line `node --inspect-brk last_javascript.mjs` and then open up `chrome://inspect` in chrome based browers and debug the javascript.

### DIFF
--- a/lua/rest-nvim/script/javascript.lua
+++ b/lua/rest-nvim/script/javascript.lua
@@ -1,0 +1,159 @@
+local script = {}
+local logger = require("rest-nvim.logger")
+
+local has_js = vim.fn.executable('node')
+
+if has_js == 0 then
+    print("node is not available")
+end
+
+local function cwd()
+  local current_script_path = debug.getinfo(1).source:match("@(.*)")
+  return current_script_path:match("(.*[/\\])") or "."
+end
+
+local function read_file(filename)
+    local file_to_open = cwd() .. "/" .. filename
+
+    local file = io.open(file_to_open, "r")
+    if file then
+      local file_content = file:read("*all")
+      file:close()
+      return file_content
+    else
+      print("Error: Could not open file " .. file_to_open)
+    end
+end
+
+local function write_file(filename, content)
+    local file_to_open = cwd() .. "/" .. filename
+    local file = io.open(file_to_open, "w")
+    if file then
+      file:write(content)
+      file:close()
+    else
+      print("Error: Could not open file " .. file_to_open)
+    end
+    return file_to_open
+end
+
+local function local_vars (ctx)
+    local ctx_vars = {}
+    for k, v in pairs(ctx.vars) do
+        ctx_vars[k] = v
+    end
+    for k, v in pairs(ctx.lv) do
+        ctx_vars[k] = v
+    end
+    return ctx_vars
+end
+
+local function create_prescript_env(ctx)
+    return {
+        _env = { cwd = cwd() },
+        request = { variables = local_vars(ctx) }
+    }
+end
+
+local function create_handler_env(ctx, res)
+    local response = res
+    -- TODO check mime type before parsing
+    local ok, decoded_body = pcall(vim.fn.json_decode, res.body)
+    if ok then
+      response = vim.deepcopy(res)
+      response.body = decoded_body
+    end
+
+    return {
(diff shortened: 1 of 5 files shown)


lua/rest-nvim/script/javascript.lua
```diff
@@ -0,0 +1,170 @@
+local script = {}
+local logger = require("rest-nvim.logger")
+
+local has_js = vim.fn.has('node')
```
boltlessengineer, line 4: `has()` is for checking features, not executables. It will check if Neovim node provider is loaded (`:h node-client`, `:h g:loaded_node_provider`) and running `node` is still possible when node provider is not loaded.

We can use `vim.fn.executable('node')` instead. (`:h executable()`)

lua/rest-nvim/script/javascript.lua
```diff
@@ -0,0 +1,170 @@
...
+      local file_content = file:read("*all")
+      file:close()
+      return file_content
+    else
+      print("Error: Could not open file " .. file_to_open)
+    end
+end
+
+local function write_file(filename, content)
+    local file_to_open = cwd() .. "/" .. filename
```
boltlessengineer, line 29: if script file is saved for temporary use, it should be saved in temporary directory like `$XDG_RUNTIME_DIR` where you can find with `stdpath("run")`

lua/rest-nvim/script/javascript.lua
```diff
@@ -0,0 +1,170 @@
...
+    return {
+        _env = { cwd = cwd() },
+        request = { variables = local_vars(ctx) }
+    }
+end
+
+local function create_handler_env(ctx, res)
+    local env_vars = {}
+    -- can't do pairs(vim.env) instead need to use environ
+    for k, v in pairs(vim.fn.environ()) do
```
boltlessengineer, line 61: I'm not sure if passing environment variables is necessary. `vim.env` table will be available as normal environment variables in processes under Neovim instance.

lua/rest-nvim/script/javascript.lua
```diff
@@ -0,0 +1,170 @@
...
+        -- ignore values which will make escaping complicated
+        if not string.find(v, "[\\'\"`]") then
```
boltlessengineer, line 63: just not handling variables with quotes feel too hacky imho.

boltlessengineer (commented):
Thank you for your interest and all works!
Didn't expect someone will contribute to javascript runner.

Though I think injecting all data directly to javascript code is too hacky. Both environment variables and response data can be quite large. I'm thinking of communicating with node client using jsonrpc to share data asynchronously.

lua/rest-nvim/script/javascript.lua
mrloop: updated

lua/rest-nvim/script/javascript.lua
mrloop: Using `process.env` in node to read environment variables now

lua/rest-nvim/script/javascript.lua
mrloop: no longer copying all the things

lua/rest-nvim/script/javascript.lua
mrloop: not sure the best way to do this and still let node resolve dependencies

mrloop:
Thanks for reviewing

>  I'm thinking of communicating with node client using jsonrpc to share data asynchronously.

That sounds good. Would you still need something like `javascript.mjs` and then the jsonrpc would be used for setting the global and request variables only? Is this PR good enough in the mean time?

boltlessengineer:
I think it would be better to prioritize implementing a proper jsonrpc version from the start. This approach will take more time, but it will provide a more maintainable and appropriate solution in the long run.

For this reason, I’m closing this PR, but I truly value your work and hope you’ll continue contributing in the future. :)

mrloop:
I look forward to seeing the jsonrpc implementation

PR #519, opened 2025-02-07
feat: add ability to pass certs

NigelGreenway:
A key feature that is missing from what I can tell is the ability to pass a cert/key. This change request is to add a cert/key via configuration based on the host. If it matches that host, then the file path for the certs/keys will be added to the request.

Changes:
 - Add `--cert` and `--key` flag with file path to docs
 - Add a configuration setup so the keys are only sent with the corresponding request

### DIFF
--- a/lua/rest-nvim/client/curl/cli.lua
+++ b/lua/rest-nvim/client/curl/cli.lua
@@ -167,18 +167,29 @@ local builder = {}
 ---@return string[] args
 function builder.extras(req)
     local args = {}
+
     if config.clients.curl.opts.set_compressed then
-        if
-            vim.iter(req.headers):any(function(key, values)
-                return key == "accept-encoding"
-                    and vim.iter(values):any(function(value)
-                        return value:find("gzip")
-                    end)
+        if vim.iter(req.headers):any(function(key, values)
+            return key == "accept-encoding" and vim.iter(values):any(function(value)
+                return value:find("gzip")
             end)
-        then
+        end) then
             vim.list_extend(args, { "--compressed" })
         end
     end
+    --
+    for domain, _ in pairs(config.clients.curl.opts.certificates) do
+        local target = req.url
+
+        local s, _ = string.find(target, domain, 1, true)
+
+        if s ~= nil then
+            vim.list_extend(args, { "--cert", config.clients.curl.opts.certificates[domain].set_certificate_crt })
+            vim.list_extend(args, { "--key", config.clients.curl.opts.certificates[domain].set_certificate_key })
+            break
+        end
+    end
+
     return args
 end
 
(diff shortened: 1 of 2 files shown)


NigelGreenway:
I still need to add tests and update the docs, and sort out some code that I've amended in error with formatting. Is this something that you would be interesting in?

lua/rest-nvim/client/curl/cli.lua
boltlessengineer, line 1: Please avoid including unnecessary code reformats. It makes reviewing bit hard.

lua/rest-nvim/client/curl/cli.lua
```diff
@@ -273,7 +284,10 @@ function builder.statistics()
             return ("? %s:%%{%s}\n"):format(style.id, style.id)
         end)
         :join("")
-    return { "-w", "%{stderr}" .. format }
+
+    local output_str = "'%{stderr}" .. format .. "'"
```
boltlessengineer, line 288: I'm curious what you tried to fix with this part.

boltlessengineer (changes requested):
Thank you for your contribution! Yes, we definitely need this feature. I can merge this PR when it is ready.
Please avoid including unnecessary code reformats ;)

lua/rest-nvim/client/curl/cli.lua
NigelGreenway: Yep, I was doing this for readability on my end, but committed in error 🤦🏽‍♂️

lua/rest-nvim/client/curl/cli.lua
NigelGreenway: I thought it was an error due to lack of quotes... I'm new to Lia, so still learning the language. This will also be reverted 😉

NigelGreenway:
Please find all comments resolved, and I've added docs and tests.

Linting, formatting and tests are all green :muscle:

lua/rest-nvim/client/curl/cli.lua
```diff
@@ -179,6 +179,19 @@ function builder.extras(req)
             vim.list_extend(args, { "--compressed" })
         end
     end
+    --
+    for domain, _ in pairs(config.clients.curl.opts.certificates) do
+        local target = req.url
+
+        local s, _ = string.find(target, domain, 1, true)
```
boltlessengineer, line 186: We already have domain matching functionality in `cookie_jar.lua` so we can use same logic for both, but for now, I think this is fine. I might add a TODO comment here though.

https://github.com/rest-nvim/rest.nvim/blob/404a26220a4835fa971df03b646a36d9c6616e0b/lua/rest-nvim/cookie_jar.lua#L172-L199

boltlessengineer (approved):
LGTM. I’ll squash-merge it after adding a commit with some minor fixes.

PR #528, opened 2025-02-14
Updates readme with latest certs feature that was missed

NigelGreenway:
In my previous [PR](https://github.com/rest-nvim/rest.nvim/pull/519), I missed updating the README.md and used the incorrect default in the vim docs.

Changes:
 - Updates `README.md` doc
 - vim doc fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ local default_config = {
                 ---@type boolean Add `--compressed` argument when `Accept-Encoding` header includes
                 ---`gzip`
                 set_compressed = false,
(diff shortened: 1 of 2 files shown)


boltlessengineer:
For missing documents in `doc/`, I'm aware of it. I'll update it soon.
The doc should not be updated manually but rather auto-generated using `sync-readme` or `docgen` binaries shipped in `flake.nix`.

For example config, I'd rather use wiki later, but thank you for notifying me. It's always better to have real-world examples.

PR #540, opened 2025-03-04
Fix Statistics syntax highlight

cristianradulescu:
The Statistics syntax highlight is broken due to missing "http_stat" treesitter language. Not starting treesitter when the language is missing will fallback to the "http_stat" vim syntax defined in the project.

Before:
before

After:
after


boltlessengineer:

PR #541, opened 2025-03-05
feat: add default = true for highlight groups definition.

xieyonn:
1. Add `default = true` for highlight groups, allow user to custom style.
2. Add descriptions for highlight groups in doc `rest-nvim.txt`.

### DIFF
--- a/doc/rest-nvim.txt
+++ b/doc/rest-nvim.txt
@@ -6,7 +6,7 @@ rest.nvim                                                            *rest-nvim*
 
 
 ==============================================================================
-Table of Contents                                               *rocks-contents*
+Table of Contents                                                *rest-contents*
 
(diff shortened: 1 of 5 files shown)


boltlessengineer (changes requested):
Although your documentation looks completely fine, it will be removed when auto-generating the help pages. So we should put that inside lua codes.
You can do this with [`docgen` command](https://github.com/rest-nvim/rest.nvim/blob/949b585807cc315337602b7824ce60d0aef58bbd/nix/test-overlay.nix#L31).

I will modify your PR if you are not familiar with nix in this weekend.

xieyonn:
Thank you for your explaination! 
I'm not familiar with nix, it will be great if you can help!

boltlessengineer (approved):

PR #546, opened 2025-06-08
feat: allow $dotenv tags in variables

bulkan:
This is essentially the same change as in this commit https://github.com/rest-nvim/rest.nvim/commit/9800b593034770fee5eb460ff0e8f0b13de83300

With this change,  [in place variables](https://www.jetbrains.com/help/idea/exploring-http-syntax.html#in-place-variables) can be defined like this

```http
@token={{$dotenv TOKEN}} 
```


boltlessengineer:

PR #566, opened 2026-03-04
feat: Add support for user-defined comment directives

Otard95:
Adds a `custom_directives` config option that lets users register handlers for `# @directive` comments.

This would be a table of keyed functions where the key is the directive and the function gets called with the space separated arguments that follow.

For example, given:
```
custom_directives = { var = function(ctx, name, value) ctx:set_local(name, value) end }
```
`# @var foo bar` would be equivalent to `@foo = bar`.

Personally I'd find this very useful for securely getting values out of my password store (example below). But I see this being quite useful in general. And as you can see the implementation is really quite simple.
Of course, you could technically achieve something similar with a post-request script. This feature is more so to handle the case where you want to do mostly the same action many times or in many places, at that point copy-pasting the script around becomes very verbose and not to mention hard to maintain.

Password store example:
```lua
vim.g.rest_nvim = {
  custom_directives = {
    pass = function(ctx, variable, pass_name)
      if not (variable and pass_name) then
        print('Directive @pass expects exactly 2 arguments')
        return
      end

      local cmd = { 'pass', 'show', pass_name }
      local result = vim.system(cmd, { text = true }):wait()

      if result.code ~= 0 then
        print('Failed to get password \'' .. pass_name .. '\': ' ..
          vim.trim(result.stdout) .. vim.trim(result.stderr))
        return
      end

      ctx:set_local(variable, vim.trim(result.stdout))
    end
  }
}
```
Then:
```http
# @pass token secret/api/key
GET https://api.example.com
Authorization: Bearer {{token}}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ vim.g.rest_nvim = {
 local default_config = {
     ---@type table<string, fun():string> Table of custom dynamic variables
     custom_dynamic_variables = {},
+    ---@type table<string, rest.Opts.DirectiveFn> Table of custom directives
+    custom_directives = {},
     ---@class rest.Config.Request
     request = {
         ---@type boolean Skip SSL verification, useful for unknown certificates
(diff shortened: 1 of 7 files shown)


Otard95:
I suppose one consideration should be: Should we reimplement the `@prompt` directive as a default to this option?
This would allow users to override it, which may or may not be desirable.

